PR #19, opened 2015-10-14
Finished the targeted feature detection

dyrlund:


msdk-featuredetection/msdk-targeteddetection/src/main/java/io/github/msdk/featuredetection/targeteddetection/TargetedDetectionMethod.java
```diff
@@ -164,14 +173,25 @@ public TargetedDetectionMethod(@Nonnull List<IonAnnotation> ionAnnotations,
                 chromatogram.setDataPoints(newDataPoints);
             }
```
dyrlund: How is the best way to define the m/z value for a chromatogram? Here, I'm using the average value of all non-zero data points used to generate the chromatogram.

msdk-featuredetection/msdk-targeteddetection/src/main/java/io/github/msdk/featuredetection/targeteddetection/TargetedDetectionMethod.java
```diff
@@ -164,14 +173,25 @@ public TargetedDetectionMethod(@Nonnull List<IonAnnotation> ionAnnotations,
...
+            for(int i=0; i < buildingChromatogram.getSize() ; i++) {
+                if (mzValues[i] > 0) {
+                    values++;
+                    sum = sum + mzValues[i];
+                }
+            }
+            double newMz = sum/values;
+            chromatogram.setMz(newMz);
+
             // Add the chromatogram to the chromatogram list
```
dyrlund: How is the best way to define the m/z value for a chromatogram? Here, I'm using the average value of all non-zero data points used to generate the chromatogram.

msdk-datamodel/src/main/java/io/github/msdk/datamodel/impl/SimpleChromatogramDataPointList.java
```diff
@@ -124,7 +124,7 @@ public void setBuffers(@Nonnull ChromatographyInfo[] rtBuffer,
 
         // Check if the RT array is properly sorted
         for (int pos = 1; pos < newSize; pos++) {
-            if (rtBuffer[pos].compareTo(rtBuffer[pos - 1]) >= 0)
+            if (rtBuffer[pos - 1].compareTo(rtBuffer[pos]) >= 0)
                 throw (new MSDKRuntimeException(
```
dyrlund: Isn't it right that it has to be the comparison has to be changed?

msdk-featuredetection/msdk-targeteddetection/src/main/java/io/github/msdk/featuredetection/targeteddetection/TargetedDetectionMethod.java
tomas-pluskal: In MZmine 2 we used a median of the m/z values, so we should probably stick to that.
It might be a good idea to add a utility class that will calculate it for you, when given the chromatogram data points.

msdk-datamodel/src/main/java/io/github/msdk/datamodel/impl/SimpleChromatogramDataPointList.java
tomas-pluskal: I agree. Also, the comparison should be > and not >=.
(Alternatively, you can use the previous comparison and change the operator to < )

msdk-featuredetection/msdk-targeteddetection/src/main/java/io/github/msdk/featuredetection/targeteddetection/TargetedDetectionMethod.java
dyrlund: I think it might make sense to re-evaluate how the m/z values are calculated. The problem with a median value is:
- it doesn’t average the variation measured by the instrument
- it is sensitive to long peak tailing

Here are two solutions which could improve the m/z calculation:
- average m/z based on weighing of the intensities
- average m/z of all data points above the FWHM

I know that changing the way we calculate the m/z value will give different results compared to MZmine 2. But improving the m/z calculation should have a positive effect on:
- ppm values
- alignment of features
- identification of features when based on m/z values

What do you think?

tomas-pluskal: That could be true. But it would be good to make a bit of evaluation before doing such change. Also, try discussing this with Matej - I think the original median concept probably came from him.

nilshoffmann: Hi,
I think that both approaches make sense in different situations.
Generally speaking, a median based approach will be less sensitive
towards outliers.
Depending on the instrument's detector, its precision, and the method
used for binning/pre-processing
of the m/z values used for the single-ion chromatogram, the median
approach will result in an m/z value
that represents the larger number of m/z's in the sample, while the mean
approach will be distorted as soon
as binning has placed m/z's into adjacent bins or has moved them to the
current bin. You can make the mean more
robust by omitting the bottom and top k % of m/z values after sorting. A
related approach for preprocessing is taken
by XCMS/Camera, if I recall correctly.

To evaluate this for data from different vendors and different
preprocessing strategies, a first step would be
to observe the ratio of median and mean. Additionally, there are usually
characteristic distributions of m/z's for
specific instruments, although they may even vary over the detector's
range, so it is a classic tradeoff between
effort required and observable effect on the data quality.

Nils

Am 15.10.15 um 15:16 schrieb Tomáš Pluskal:

dyrlund:
I have moved the m/z calculation to the utility class and added three different ways to calculate the m/z values. Here is an initial test on QTOF data for 10 internal standards. The results are from one sample:

mz calculation

msdk-datamodel/src/main/java/io/github/msdk/datamodel/impl/SimpleChromatogramDataPointList.java
dyrlund: I have corrected this.

tomas-pluskal:
Looks good. On this data set the average looks like the best approach, but it might be instrument-dependent. Anyway, it's good to have a choice.

PR #20, opened 2015-10-19
Chromatogram

dyrlund:
Should the various functions have the chromatogram as input or should the input be the RT and intensity buffers? The down side with using the chromatogram is that the data points will have to be loaded each time a value has to be calculated, a step which can be slow.


tomas-pluskal:
I'd go for list of Chromatograms as input - it's cleaner and the user has a choice how the data points are stored (by providing his choice of DataPointStore).

PR #36, opened 2015-11-28
Cropfilter

SandraCastilloPriego:
All comments addressed except:
rtRange has already the nonnull tag, and getDataPointsByMzAndIntensity(dataPoints) doesn't work for me for some reason..


tomas-pluskal:
Thanks Sandra!
Just FYI, you don't need to create a new pull request to make changes to the old one. You can simply add new commits to the same branch and push the branch to your fork on github. The commits are then automatically added to the pull request.

As for the additional tests, I mean for example setting the mzRange to 2000-3000 while the raw data file has scans only 300-600 m/z. The result should be an empty RawDataFile. Another useful test could be setting the m/z range to Range.all() - the resulting RawDataFile should be identical to the old one, same number of data points etc. Having such tests is very useful in the long-term. Of course, the crop filter is not the most complicated module, but for more complex algorithms this is really good to have. Again, just FYI.

PR #38, opened 2015-11-29
Simplified ColumnName handling, moved release-related plugins to separate profile. Added new module for ChromaTOF CSV file support.

nilshoffmann:


tomas-pluskal:
THanks @nilshoffmann ! Looks good to me.
Thomas @dyrlund , can you check the ChromaTOF CSF module, if you have any comments to it?

msdk-datamodel/src/main/java/io/github/msdk/datamodel/featuretables/ColumnName.java
```diff
@@ -28,7 +26,7 @@
     GROUPID("Group ID", Integer.class),
     MZ("m/z", Double.class),
     PPM("ppm", Double.class),
-    RT("RT", ChromatographyInfo.class),
+    RT("RT", Double.class),
     RTSTART("RT Start", Double.class),
```
dyrlund: Why is the class of RT changed to Double from ChromatographyInfo?

tomas-pluskal: Hm, should be ChromatographyInfo, I think.

dyrlund:
@tomas-pluskal I'm not sure I understand your comment about checking the ChromatTOF CSF module. The import module is still under development by Nils.

tomas-pluskal:
Yes, sorry, I wrote that before I actually checked the code :)

msdk-datamodel/src/main/java/io/github/msdk/datamodel/featuretables/ColumnName.java
nilshoffmann: Thanks for spotting this, I will check where I need to fix this. There was an issue with one of the tests, when it was ChromatographyInfo. Maybe on of the builders was not setting the type correctly. Will fix this with the next pull request.

nilshoffmann:
Yes, I only added it as a stub for now. More will follow soon.

PR #44, opened 2015-11-30
Updated source and target to Java 1.8

nilshoffmann:
Hi, 
I am not sure, whether the proposed simplification regarding the compiler plugin configuration works with the Maven Builder in Eclipse. It does work in NetBeans and with Maven via the CLI. 
Did we actually want to target Java 1.8? If not, I will update the pull request.
I have added the JaCoCo plugin for code coverage, activated when the -Pjacoco profile is set.


tomas-pluskal:
Hi Nils,
The target platform for MSDK is Java 7.

nilshoffmann:
Okay, I have changed the version to 1.7. I also extracted some of the versions into properties for simpler maintenance / updates.

PR #45, opened 2015-12-01
Finished reorganizing MSDK modules, created modules for metfrag, sirius

tomas-pluskal:


tomas-pluskal:
Hi Thomas @dyrlund , Nils @nilshoffmann , Sandra @SandraCastilloPriego , Gert @berlinguyinca 

I added new modules to MSDK for additional methods like MetFrag and SIRIUS. Since the number of modules has increased quite a bit and there was no naming scheme so far, I renamed the existing modules to roughly follow a naming pattern msdk-area-modulename. Unless someone objects, I think we can keep this naming pattern stable for a while..

screen shot 2015-12-01 at 08 00 08

PR #50, opened 2015-12-02
Mean scan filter method

SandraCastilloPriego:
I added a new folder for the scan filters inside the rawdata-filters module. I've added the MeanFilterMethod.


msdk-rawdata/msdk-rawdata-filters/src/test/java/io/github/msdk/filtering/scanfilters/MeanFilterMethodTest.java
```diff
@@ -0,0 +1,126 @@
...
+        Assert.assertEquals(1.0, meanFilter.getFinishedPercentage(), 0.0001);
+
+        // The result of the method can't be Null
+        Assert.assertNotNull(newScan);
+
+        // The resulting scan should be equal to the input scan
+        MsSpectrumDataPointList dataPoints = MSDKObjectBuilder.getMsSpectrumDataPointList();
+        scanTest.getDataPoints(dataPoints);
+        List<Ion> ions = new ArrayList();
+        List<Ion> newIons = new ArrayList();
```
tomas-pluskal: What is the purpose of using ArrayList here?
You could just make a new MsSpectrumDataPointList, and copy the contents by MsSpectrumDataPointList.copyFrom(dataPoints).. no?

SandraCastilloPriego: It was because I'm checking their intensity and mz values one by one and I wanted to be sure I got the same value using the same index. 
Is it better to use the iterators? Something like this?
while(interator1.hasNext()){
    MsIon ion1 = iterator1.next();
    MsIon ion2 = iterator2.next();
}

tomas-pluskal: I'd suggest to use the arrays instead of iterators, e.g.
```
double mzValues1[] = dataPoints1.getMzBuffer();
double mzValues2[] = dataPoints2.getMzBuffer();
for (int i = 0; i < dataPoints1.getSize(); i++) {
   ... compare mzValues1[i] vs mzValues2[i]
}
```

msdk-rawdata/msdk-rawdata-filters/src/test/java/io/github/msdk/filtering/scanfilters/MeanFilterMethodTest.java
```diff
@@ -0,0 +1,110 @@
...
+        
+        // Get the mz and intensities values from the input scan
+        double mzValues[] = dataPoints.getMzBuffer();
+        float intensityValues[] = dataPoints.getIntensityBuffer();
+        int numberOfDataPoints = dataPoints.getSize();
+        
+        newScan.getDataPoints(dataPoints);
+        
+        // Get the mz and intensities values from the filtered scan
+        double newMzValues[] = dataPoints.getMzBuffer();
```
tomas-pluskal: This call to dataPoints.getMzBuffer() is going to return the same double[] array as the previous call a few lines above - so your comparison will always be true. You will need to create two separate MsSpectrumDataPointList instances, one for the original data points and one for the filtered ones.

SandraCastilloPriego:
I've realized it doesn't pass the last test and I can't see why. The average of all the intensities in the input is not exactly the same as the value of all the data points in the new filtered scan but it should be...

tomas-pluskal:
It might be easier to discover the cause of the problem if you use a smaller test case (e.g., make an instance of MsScan with only 10 manually defined data points).

tomas-pluskal:
Sandra, a general comment to the filtering methods: I think people might want to use the MSDK methods in various ways. One way is to filter the whole RawDataFile and produce a new RawDataFile. Another way is to filter a single spectrum (for whatever purpose they need). So I think it might be worth to split your methods into 2 classes - one that will take an MsSpectrum as an input and generate a new, filtered MsSpectrum (or null if the spectrum is to be removed completely). This class will contain the actual logic of the algorithm. I think it does not need to implement MSDKMethod, but maybe we can add an interface like MSDKFilteringAlgorithm
```
public interface MSDKFilteringAlgorithm {
   @Nullable MsSpectrum performFilter(@Nonnull MsSpectrum input) {
     ...
 }
}
```
The second class can provide the full-scale MSDKMethod to filter the whole RawDataFile according to the selected algorithm. E.g., something like this:
```
public class MSDKFilteringMethod implements MSDKMethod<RawDataFile) {
   public MSDKFilteringMethod(RawDataFile inputFile, MSDKFilteringAlgorithm filteringAlgorithm) {
    ...
   }
}
```
This approach would allow more flexible ways to use the algorithms. What do you think?

SandraCastilloPriego:
Should I do the same for the crop filter?

SandraCastilloPriego:
Where should I create the Algorithm interface?

tomas-pluskal:
What I'm thinking is to have only one filtering Method, and many Algorithms that this method can apply. 
You can keep the Algorithm interface in the same maven package.

SandraCastilloPriego:
Is it something like this you were thinking?

tomas-pluskal:
Yes, exactly. 
But the casting from MsSpectrum to MsScan is potentially dangerous, so it might be better to change the method signature in MSDKFilteringAlgorithm to 
```
public @Nullable MsScan performFilter(@Nonnull MsScan scan) {
   ...
}
```

msdk-rawdata/msdk-rawdata-filters/src/main/java/io/github/msdk/filtering/MSDKFilteringMethod.java
```diff
@@ -1,116 +1,92 @@
...
+        }
+    }
+
+    @Override
+    public RawDataFile execute() throws MSDKException {
+        logger.info("Started Filter " + this.filteringAlgorithm.getName() + " with Raw Data File #"
+                + rawDataFile.getName());
+
+        // Create a new raw data file
+        result = MSDKObjectBuilder.getRawDataFile(this.rawDataFile.getName(), this.rawDataFile.getOriginalFile(), this.rawDataFile.getRawDataFileType(), DataPointStoreFactory.getMemoryDataStore());
```
tomas-pluskal: The data point store should be a parameter of this method (in constructor).
If you hardcode in ```DataPointStoreFactory.getMemoryDataStore()``` then MSDK will run out of memory very quickly.

PR #61, opened 2015-12-11
Added MS/MS feature detection method

dyrlund:


tomas-pluskal:
Hi Thomas,
I think the ScanSelection class does not really belong to MSDK, it's more MZmine-oriented. Perhaps this method should just take a List<MsScan> as its input, and not deal with the way how the scans are selected.

dyrlund:
Sounds like a good idea - I have made the changes.

PR #65, opened 2015-12-15
Added the local database search identification module

dyrlund:


msdk-datamodel/src/main/java/io/github/msdk/util/IonTypeUtil.java
```diff
@@ -0,0 +1,88 @@
...
+        } catch (Exception e) {
+            numberOfMolecules = 1;
+        }
+
+        // Adduct formula
+        String adductFormula = null;
+        /*
+         * TODO: Extract e.g. NH4 from [M+NH4]+
+         */
+        adductFormula = "[M+H]";
```
dyrlund: @tomas-pluskal Can you help with an easy way to extract the adduct information from the string? My RegExp skills are not very strong.

PR #66, opened 2015-12-15
Various updates

tomas-pluskal:


msdk-datamodel/src/main/java/io/github/msdk/util/IonTypeUtil.java
```diff
@@ -28,6 +36,9 @@
  */
 public class IonTypeUtil {
 
+    private static final Pattern ionTypePattern = Pattern
+            .compile("\\[(\\d*)M([+-]?.*)\\](\\d*)([+-])");
+
     /**
```
dyrlund: @tomas-pluskal Adducts are matched as +NH4 not NH4. Can you help with fixing this?

PR #69, opened 2015-12-17
Adding tests to the ChromatogramBuilderMethod.

SandraCastilloPriego:
Hi, 
I added few tests but the method doesn't pass the last test. When the MZTolerance is 0.0 it shouldn't be any chromatrogram. Am I wrong?

### DIFF
--- a/msdk-featdet/msdk-featdet-chromatogrambuilder/src/test/java/io/github/msdk/featuredetection/chromatogrambuilder/ChromatogramBuilderMethodTest.java
+++ b/msdk-featdet/msdk-featdet-chromatogrambuilder/src/test/java/io/github/msdk/featuredetection/chromatogrambuilder/ChromatogramBuilderMethodTest.java
@@ -47,18 +47,56 @@ public class ChromatogramBuilderMethodTest {
         Assert.assertNotNull(rawFile);
         Assert.assertEquals(1.0, importer.getFinishedPercentage(), 0.0001);
 
-        final double minimumTimeSpan = 6; // 6s
-        final double minimumHeight = 1E4;
-        final MZTolerance mzTolerance = new MZTolerance(0.001, 5.0);
+        double minimumTimeSpan = 6; // 6s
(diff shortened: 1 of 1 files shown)


dyrlund:
@tomas-pluskal Did you ever finished the chromatogram builder method?

tomas-pluskal:
I finished it, but never had the time to test it. So I think it's great that Sandra is adding tests and I can debug the issues.

dyrlund:
@tomas-pluskal I added the chromatogram builder test to MZmine 3 and it runs very slow. Building chromatograms for a single 50 MB mzML file takes about 30 minutes.

tomas-pluskal:
Did you run it on profile or centroid data? 
Since we abolished "mass lists", the chromatogram builder must be run on centroid data.
We already have centroiding modules in MSDK, I am going to add them to MZmine 3 menu.

dyrlund:
It was centroid data. The mzML file contains about 2500 MS1 scans.

tomas-pluskal:
Can you send me a copy of the file and the parameters you used? I would like to test it.

dyrlund:
I have send you the file through mailbigfile.com, you can also download it here: mbf.me/pDgyCs

**Import**
File location: Desktop on C-drive
Import mode: Store the file contents in memory

**Chromatogram builder**
m/z tolerance: 0.005 m/z or 5.0 ppm
Min peak duration: 5.00
Min peak height: 5.00E3
Name suffix: " chromatograms"

The processing is done on a Windows 10 Pro 64-bit, i7-4790S CPU with 16 GB of memory and takes about 25 minutes to complete. A total of 962 features are found.

tomas-pluskal:
Hi Thomas,
Can you try the chromatogram builder again now that it includes the "noise level" option?
Please set the noise level to the same number you would use for the Centroid mass detector in MZmine 2.

dyrlund:
Much better. With a noise level of 500 the file is now processed in 4 minutes!

PR #87, opened 2016-01-01
Minor fixes after minor fixes in the data model.

nilshoffmann:
The automatic setting of the scan's file broke a number of tests which are fixed by this commit. Please check whether the change on lines 66-68 in  msdk-datamodel/src/main/java/io/github/msdk/util/MsScanUtil.java is acceptable (commented out for now).


tomas-pluskal:
Thanks, Nils!

PR #90, opened 2016-01-04
Added an interface for spectral comparison algorithms

tomas-pluskal:
Hi Gert, Happy New Year!
How about adding the spectral comparison algorithms to MSDK, as we discussed at Dagstuhl?
I added a simple interface for these methods.


tomas-pluskal:
Gert @berlinguyinca , are you there?

berlinguyinca:
Sorry for the late response,  have been busy with some other stuff and
sounds good.

PR #95, opened 2016-01-24
Introduced msdk-release profile to avoid CI build failures.

nilshoffmann:


tomas-pluskal:
Looks good, Nils. Thanks!

PR #102, opened 2016-02-15
Renamed Method to Algorithm, following MSDK convention

tomas-pluskal:


dyrlund:
Will you make this change throughout MSDK for all the algorithms?

tomas-pluskal:
It is done, I believe. The convention is that Algorithm are classes with fast tools implemented as static methods, so they're quick to use and we don't need to track their progress. On the other hand, Methods implement MSDKMethod and track progress.

dyrlund:
Thanks for the explanation - I though that all existing modules were going to be refereed to as algorithms.

PR #105, opened 2016-02-16
Added CsvFileExportMethod

dyrlund:


msdk-io/msdk-io-csv/src/main/java/io/github/msdk/io/csv/CsvFileExportMethod.java
```diff
@@ -0,0 +1,264 @@
...
+                        strValue = strValue + itemSeparator;
+                    }
+                    if (rt2 != null) {
+                        strValue = strValue + rt2.toString();
+                    }
+                }
+                // Ion annotations
+                else if (object instanceof List<?>) {
+                    List<IonAnnotation> ionAnnotations = (List<IonAnnotation>) object;
+                    for (IonAnnotation ionAnnotation : ionAnnotations) {
```
dyrlund: I can't find a way to check if the object is a List<IonAnnotation> object. Any suggestions?

tomas-pluskal: That is not possible, because java generics are recognized only at compilation time, not at runtime. At runtime, all Lists are the same, no matter if they were declared as List<X> or List<Y>.

It is probably safer to iterate over the list and make sure each object you get is actually instanceof IonAnnotation.

PR #110, opened 2016-02-19
Added a test for the MzTabFileExportMethod

dyrlund:


msdk-io/msdk-io-mztab/src/test/java/io/github/msdk/io/mztab/MzTabFileExportMethodTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        // RT
+        FeatureTableColumn<ChromatographyInfo> chromInfoColumn = featureTable2
+                .getColumn("Chromatography Info", null,
+                        ChromatographyInfo.class);
+        ChromatographyInfo chromatographyInfo = row.getData(chromInfoColumn);
+        Assert.assertNotNull(chromatographyInfo);
+        Float rt = chromatographyInfo.getRetentionTime();
+        Assert.assertNotNull(rt);
+        Assert.assertEquals(30.324697494506836, rt, 0.0000001);
```
dyrlund: The test fails when checking the RT value. Because the expected value is 30.324697494506836 and the test value is 30.259201049804688. 

The `MzTabFileImportMethodTest `performs the same test which completes fine so I guess there is something wrong with handling the RT value in the `MzTabFileExportMethod`.

Can you see what I have done wrong?

tomas-pluskal: What is the exact value in the exported mzTab file?

dyrlund: I can see that it is 30.324697380952383. I thought it was 30.324697494506836 which might indicate there is also an issue with the mzTab import.

tomas-pluskal: The difference between 30.324697380952383  and 30.324697494506836 is probably just a precision limit of Float.

dyrlund: After doing some more debugging, I found that the error was caused because `sm.setRetentionTime(rtValue);` was run before `sm.setOptionColumnValue(new Assay(sampleCounter), "rt", rtValue);`. This resulted in the common RT value being set to the value of the last sample. The order has therefore been changed in #112.

Issue #108, opened 2016-02-17
Expanding the IonAnnotation based on the mzTab format


dyrlund:
In the current mzTab format (v. 1.0.0) there are mandatory columns for small molecules which are currently not supported in the `IonAnnotation` - see page 13 in the [mzTab Specifications](https://mztab.googlecode.com/svn/specification_document-releases/version-1.0.0/mzTab_format_specification.pdf). These are:
- inchi_key
- taxid
- species
- database
- database version
- spectra_ref
- search_engine
- best_search_engine_score
- modifications
- reliability

I think this information will make it more transparent where identifications are coming from when using multiple identification methods. Furthermore, it will also provide some of the structure needed to implement MS/MS identification methods.

What do you think about adding support for this in MSDK?

tomas-pluskal:
I agree it is useful.

dyrlund:
We should also add the following two:

- smiles
- inchi_key

dyrlund:
Based on the mzTab specifications, I suggest to use the following class types:

- smiles - String
- inchi_key - String
- taxid - Integer
- species - String
- database - String
- database_version - String
- spectra_ref - String
- search_engine - String
- best_search_engine_score - Double
- modifications - String
- reliability - Integer

Any objections against that? And any suggestions about how we best make the change in MSDK?

Adding them all seperately to the IonAnnotation is the simplest but this will not take into account that the following are related:
- taxid and species
- database and database_version

dyrlund:
@tomas-pluskal Any comments about how we best make the changes in MSDK?

dyrlund:
I have made the changes as suggested above in #121. Please let me know what you think.

PR #121, opened 2016-03-03
Updated the IonAnnotation model to include more information

dyrlund:
The changes are bassed on issue #108.

Please note that the updates to the IonAnnotation also required adding the new methods to the MonaSpectrum. These methods are however auto-generated method stub.


tomas-pluskal:
Thanks, I think this is nice for now, later we can perhaps add more options to the CSV export module (e.g., whether to export all columns or only columns that have some data).

PR #122, opened 2016-03-03
Initial work on the SRM feature detection module

dyrlund:


tomas-pluskal:
I was just thinking we need this, cause I started running some SRM acquisitions :)

PR #123, opened 2016-03-04
Updates to the SRM feature detection module

dyrlund:


dyrlund:
My idea with the SRM feature detection module is to first find all the highest data points of the possible peaks in each SRM chromatogram. These data points will then be used to generate the XIC profiles as we currently do in the targeted detection method. Any suggestions to this?

As this SRM chromatogram shows, each chromatogram can include multiple peaks at different retention times which is why we can't assume only one peak in each chromatogram:
chromatogram

tomas-pluskal:
Hi Thomas,

I think the SRM detector should just build chromatograms, that's all (similar to the standard Chromatogram builder). The individual peaks in the chromatograms can later be separated using the Chromatogram deconvolution module.

dyrlund:
Good point - I think your suggestion makes more sense and then the deconvolution module can separate the individual peaks. I'll make the change.

PR #132, opened 2016-03-21
Adding initial module for feature normalization by internal standards

dyrlund:
@tomas-pluskal I'm not sure how we best name the normalization methods. If you have a better suggestion then please feel free to change the names or how the modules are located.

@Ash706 As agreed, here is the initial template for the normalization module. I have added a test mzTab file which you can use.


Ash706:
Thanks for adding the initial module and the test file. I hope to take it forward.

PR #144, opened 2016-09-16
MzTabFileImportMethod: check if MZTabFile is null

TDI-tenderholt:
This fixes the ```NullPointerException``` error in ```MzTabFileExportMethodTest``` (see https://travis-ci.org/msdk/msdk/builds/150175189#L6132), although the test now fails instead. So far, I have been unable to track down that problem which I assume is in ```MzTabFileExportMethod```.

### DIFF
--- a/msdk-io/msdk-io-mztab/src/main/java/io/github/msdk/io/mztab/MzTabFileExportMethod.java
+++ b/msdk-io/msdk-io-mztab/src/main/java/io/github/msdk/io/mztab/MzTabFileExportMethod.java
@@ -150,8 +150,8 @@ public class MzTabFileExportMethod implements MSDKMethod<File> {
         mtd.addVariableModParam(1, new CVParam("MS", "MS:1002454",
                 "No variable modifications searched", null));
 
-        // Create stable columns - only available in jmztab 3.0.2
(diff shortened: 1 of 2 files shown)


tomas-pluskal:
I think this is an issue in jmztab. I reported it a while ago:
https://github.com/PRIDE-Utilities/jmzTab/issues/4
But nobody responded there...

TDI-tenderholt:
Yes, odd that you haven't seen a result for the jmztab issue.

Given that it's possible to get a null from problematic jmztab, doesn't it make sense to at least check for the null instead of allowing an exception?

Btw, as expected, the CI check is now a failure instead of error (https://travis-ci.org/msdk/msdk/builds/160447602#L6138).

TDI-tenderholt:
So it looks like uncommenting [```factory.addDefaultStableColumns()```](https://github.com/msdk/msdk/blob/master/msdk-io/msdk-io-mztab/src/main/java/io/github/msdk/io/mztab/MzTabFileExportMethod.java#L154) fixes the failure.

If you plan on requiring jmztab 3.0.2+, then this seems like a legit fix.

tomas-pluskal:
Looks good, thanks a lot for digging into this!

PR #145, opened 2016-09-20
Convert MZTolerance to an interface

TDI-tenderholt:
* This will allow different ways of calculating a m/z tolerance (e.g. constant m/z, constant ppm, variable m/z or ppm, m/z-specific, etc). The primary motivation is to support error bars for each peak found by PeakInvestigator v1.4, but one could imagine other uses:

```
MaximumTolerance(MZTolerance[] tolerances)   // get maximum of all tolerances
MSLevelTolerance(MsScan scan)                // get m/z tolerance based on ms-level
...
```

* The original ```MZTolerance``` class has been renamed ```CombinedMZTolerance```, but should probably be converted to ```MaximumTolerance``` above.
* Doc strings still need to be fixed to specify that ```MZTolerance``` is now an interface.


tomas-pluskal:
Hi Adam,
Looks fine to me.
I agree with renaming CombinedMZTolerance to MaximumTolerance (perhaps MaximumMZTolerance).
Also, could you add the license header to MZTolerance.java?

TDI-tenderholt:
Sure, I can add the license header to ```MZTolerance.java```. I'll also try to update all docstrings that mention ```MZTolerance```.

```MaximumMZTolerance``` works for me. Do you prefer that capitalization scheme, even though ```MaximumMzTolerance``` follows a common [Java standard](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case)?

Finally, are you ok with these going into ```io.github.msdk.util``` or would you like another package? I think there will be the ```MZTolerance``` interface, and at least a few class implementations like ```ConstantPPMTolerance```, ```ConstantDaTolerance```, ```MaximumMZTolerance```.

**Comment by tomas-pluskal** [comment]
You're right, MaximumMzTolerance is better.

Since we have a few classes now, perhaps we can move them to a new package, like io.github.msdk.util.tolerances?

**Comment by TDI-tenderholt** [comment]
Ok, almost there.

I created a new general ```MaximumMzTolerance``` class that accepts 2 or more ```MzTolerance``` objects, but I'm not sure how to apply it to [```GapFillingMethod```](https://github.com/msdk/msdk/blob/master/msdk-features/msdk-features-gapfilling/src/main/java/io/github/msdk/features/gapfilling/GapFillingMethod.java#L172) because of the specialized getters. Do you want a ```CombinedConstantDaPpmTolerance``` class inside the tolerances subpackage since it seems to be a widely-used case?

Or is there a more elegant solution for ```GapFillingMethod```? It's not clear exactly how it works nor why increasing the m/z tolerance through each iteration through the ```totalGaps``` is optimal.

**Comment by TDI-tenderholt** [comment]
Actually, now that I look at ```GapFillingMethod``` a bit closer, perhaps it's a good candidate for m/z-specific tolerances. Does that make sense?

**Comment by tomas-pluskal** [comment]
Hi Adam,
I think you might be over-combining it a bit.. What I meant to do was to simply rename the previous MZTolerance class to MaximumMzTolerance (taking both ppm and Da values and calculating whichever is larger). I understand your approach with MaximumMzTolerance taking multiple MZTolerance instances, but I think it is taking the abstraction a bit a too far and somewhat impractical in real life.

**Comment by tomas-pluskal** [comment]
I think the m/z tolerance calculation in the gap filler is wrong, I filed a new issue for that and will fix it when I find some time.. (https://github.com/msdk/msdk/issues/146)

**Comment by TDI-tenderholt** [comment]
I agree that it's probably impractical to combine more than two MzTolerance instances—or even more than one of the same type (e.g. ```ConstantDaTolerance```)—but isn't it better to at least have the flexibility to support the general combination of two different MzTolerance instances instead of just constant Da and ppm?

I can't think of anything for the maximum combination, but for a minimum combination, it might be useful to combine a m/z-specific tolerance with a global constant tolerance. For example, suppose you had data with nearly isobaric species that have resolved monoisotopic peaks, but the rest of the isotopic series remained unresolved. If you were doing identification work, it would be useful to set a smaller tolerance specifically for the resolved monoisotopic peaks than for the entire spectrum.

The only immediate problem is related to gap filler, which goes away if your new solution doesn't require any getters.

Of course, if you still think it's not useful to have a general way to combine two MzTolerance instances other than the maximum of Da and ppm, I'll make the change.

**Comment by tomas-pluskal** [comment]
Hi Adam,
for now, please just comment-out these lines in the Gap Filler module, so we can merge your pull request:
```
                mzTolerance = new MZTolerance(
                        mzTolerance.getMzTolerance() + toleranceMz,
                        mzTolerance.getPpmTolerance());
```
I will fix it later.

As for the combined MzTolerance class, I don't see a need for it now, it will just make the code appear more complicated than it is. If someone decides to develop a specialized module that requires it (such as the isotope identification you mentioned), then we can add it.

TDI-tenderholt:
I've removed the over-abstracted ```MaximumMzTolerance``` class and replaced it with the original Da + ppm tolerance class as discussed above. I also changed the gap filling method to use the interface and commented out the lines as you requested.

Let me know if there is anything else.

tomas-pluskal:
Looks good, thanks!

PR #148, opened 2016-09-23
Support different MzTolerances for each scan in the chromatogram builder

TDI-tenderholt:
This is first step for having the chromatogram builder take advantage of the error bars provided by PeakInvestigator v1.4.

Second step will modifying ```ChromatogramBuilderMethod``` constructor to take a ```MzToleranceFactory``` interface instead of ```MzTolerance```. Proposed interface is:

```
interface MZToleranceFactory {
    MzTolerance getMzTolerance(MsScan scan);
}
```

### DIFF
--- a/msdk-datamodel/src/main/java/io/github/msdk/util/tolerances/ConstantDaTolerance.java
+++ b/msdk-datamodel/src/main/java/io/github/msdk/util/tolerances/ConstantDaTolerance.java
@@ -14,11 +14,13 @@
 
 package io.github.msdk.util.tolerances;
 
+import io.github.msdk.datamodel.rawdata.MsScan;
+
 import javax.annotation.Nonnull;
 
 import com.google.common.collect.Range;
 
-public class ConstantDaTolerance implements MzTolerance {
+public class ConstantDaTolerance implements MzTolerance, MzToleranceProvider {
 
     private final @Nonnull Double mzTolerance;
 
@@ -63,5 +65,10 @@ public class ConstantDaTolerance implements MzTolerance {
     @Override
     public String toString() {
         return mzTolerance + " Da";
+    }
+
+    @Override
+    public MzTolerance getMzTolerance(MsScan scan) {
+        return this;
     }
 }
--- a/msdk-datamodel/src/main/java/io/github/msdk/util/tolerances/ConstantDaTolerance.java
+++ b/msdk-datamodel/src/main/java/io/github/msdk/util/tolerances/ConstantDaTolerance.java
@@ -14,13 +14,11 @@
 
 package io.github.msdk.util.tolerances;
 
-import io.github.msdk.datamodel.rawdata.MsScan;
-
 import javax.annotation.Nonnull;
 
 import com.google.common.collect.Range;
 
-public class ConstantDaTolerance implements MzTolerance, MzToleranceProvider {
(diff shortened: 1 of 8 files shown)


tomas-pluskal:
Hi Adam,
So far it looks fine for me. I don't like so much the name MZToleranceFactory, because it is not actually a factory type (meaning an object factory), more like provider type. Perhaps we can call it MZToleranceProvider?

What is the advantage of having m/z tolerance calculated separately for each scan? In principle, the m/z error depends on the instrument, and is not going to change from scan to scan.

TDI-tenderholt:
Good point about it being more a provider than a factory. I'll plan on ```MzToleranceProvider``` interface.

PeakInvestigator v1.4 will return the mass (and intensity) error for *each peak* in each scan, essentially providing the uncertainty in a peak call. Here's an example from our development version:

snip20160824_1

As you can see, there is far more uncertainty in the shoulder than in the main peak.

My current thinking, at least for chromatogram builder, is two-fold:

1) A user won't really have to specify any tolerance because it'll be provided by PeakInvestigator, although they may need to specify a multiplier since the values are for a single standard deviation, especially if the instrument drifts a lot between scans.

2) The tighter mass tolerances may lead to "cleaner" chromatograms, that is, nearly isobaric species that elute at different times may appear as two separate chromatograms instead of a single one.

Unfortunately, I haven't been able to test either of these ideas yet, but plan to soon.

Finally, if you can think of any other applications for the mass/intensity error bars, let me know.

tomas-pluskal:
Hi Adam,

To be honest, I think the current solution where ConstantPPMTolerance implements MzToleranceProvider is not semantically correct. It makes more sense to handle this inside the ChromatogramBuilderMethod. 
I think ChromatogramBuilderMethod should have two different constructors - one that takes MzTolerance (for the case when m/z tolerance is defined equally for all scans) and one that takes MzToleranceProvider (for the case of PeakInvestigator).

TDI-tenderholt:
I agree that it's not semantically correct, but figured it was a very simple solution since it didn't require any new objects or other code changes (i.e. new constructors as you propose).

In your scenario where ```ChromatogramBuilderMethod``` has two different constructors, are you envisioning that the instance ```mzTolerance``` and ```mzToleranceProvider``` variables are set to ```null```, and then when they need to be used, there's a simple check for ```null```? For example:

```
if (mzTolerance == null) {
   mzTolerance = mzToleranceProvider.getMzTolerance(scan);
}
```

or perhaps the following is better to preserve the ```mzTolerance``` set in the constructor:

```
MzTolerance tol = mzTolerance != null ? mzTolerance : mzToleranceProvider.getMzTolerance(scan);
```

tomas-pluskal:
The change actually does require code changes, because it breaks the backwards compatibility (in other projects that rely on MSDK and use the current constructor, such as the development version of MZmine 3).

I think the best solution might be to use MzToleranceProvider internally in the ChromatogramBuilderMethod. In the existing constructor that takes MzTolerance as a parameter, you can wrap the MzTolerance in a constant MzToleranceProvider that always returns the same MzTolerance.

tomas-pluskal:
Yup, that's fine.

TDI-tenderholt:
How's this? Anything else I should do?

tomas-pluskal:
Looks good, thanks

PR #150, opened 2016-10-11
Test cases for MgfExportAlgorithm

TDI-tenderholt:
Before working on converting ```MgfExportAlgorithm``` to a ```MSDKMethod```, I figured there should be some test cases to verify the behavior is still correct.


tomas-pluskal:
Looks good, thanks!

PR #152, opened 2016-10-27
Fix some docstrings

TDI-tenderholt:
Came across a docstring that seemed outdated (i.e. references to ```IMassSpectrum```), which I've updated to ```MsSpectrum```.

The [docstring](https://github.com/msdk/msdk/blob/master/msdk-datamodel/src/main/java/io/github/msdk/datamodel/msspectra/MsSpectrum.java#L59) of ```MsSpectrum.getMzValues()``` also references ```DataPointList``` which seems outdated. I also see similar references in ```Chromatogram```, ```DataPointStore```, and ```MemoryDataPoint```.


tomas-pluskal:
Thanks, fixing documentation is always welcome and appreciated!

TDI-tenderholt:
How should the other docstrings be fixed? They reference DataPointsList
which seems outdated.

tomas-pluskal:
DataPointList was removed a while ago. The methods now return primitive arrays, so the docs can be updated accordingly (e.g., "returns an array of m/z values").

PR #154, opened 2016-10-31
Improvements to the TxtExportAlgorithm

TDI-tenderholt:
Add methods that take optional ```String``` delimiter, and make original method signatures delegate to these new methods. ```String``` was chosen over ```char``` because delimiters such as ```, ``` (comma-space) may be desired.

Also implemented tests and improved documentation.


tomas-pluskal:
Looks good, thanks!

Issue #153, opened 2016-10-27
Bug in TxtExportAlgorithm.exportSpectra() ?


TDI-tenderholt:
It appears that [`TxtExportAglorithm.exportSpectra()`](https://github.com/msdk/msdk/blob/master/msdk-io/msdk-io-txt/src/main/java/io/github/msdk/io/txt/TxtExportAlgorithm.java#L55) will concatenate the output from multiple spectra into a single file without any line break between spectra. The only way to tell there are multiple spectra is to look for discontinuities in the m/z values. Is that intentional? If so, note that will probably not work in the perhaps-unlikely general case (e.g. spectrum 1 is 200-500, spectrum 2 is 505-800, etc.).

tomas-pluskal:
True. We can add an extra line separator in between the spectra.

PR #156, opened 2016-11-05
Print new line between spectra during text export

TDI-tenderholt:
Modified both ```TxtExportAlgorithm``` and ```TxtImportAlgorithm``` to handle exporting multiple spectra in a single file to address #153. Blank lines are placed between each spectrum, and import now has function to return a collection (if expected) or just the first encountered spectrum.

Also some documentation fixes and test clean-up.


tomas-pluskal:
Awesome, thanks

PR #157, opened 2016-11-11
Add module for centroiding data with PeakInvestigator

TDI-tenderholt:
This pull request will add a module (```msdk-rawdata-peakinvestigator```) for centroiding data using PeakInvestigator. There's still a lot to do before it can be merged, but I wanted feedback before continuing.

- [x] Method to bundle scans into a tarball (```PeakInvestigatorScanBundlingMethod```)
- [ ] Method to extract mass lists from tarball (```PeakInvestigatorScanExtractingMethod``` ?)
- [ ] Method to transfer files to/from SFTP server (```PeakInvestigatorFileTransferMethod``` ?)
- [ ] Method(s) to interact with the PeakInvestigator API

If any of those methods would be of general use (e.g. ```TarExportMethod```), let me know and I'll move them to other modules.

### DIFF
--- a/msdk-all/pom.xml
+++ b/msdk-all/pom.xml
@@ -126,6 +126,12 @@
 
 		<dependency>
 			<groupId>io.github.msdk</groupId>
+			<artifactId>msdk-rawdata-peakinvestigator</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+
+		<dependency>
+			<groupId>io.github.msdk</groupId>
 			<artifactId>msdk-rawdata-xic</artifactId>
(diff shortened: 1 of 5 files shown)


tomas-pluskal:
Hi Adam,
Wouldn't it be more efficient to transfer the scan data as binary data instead of the text representation?

TDI-tenderholt:
I agree it'd be more efficient to transfer the scan data as binary, but PeakInvestigator currently only handles delimited text. It's probably a feature we should consider adding, especially if there's major benefit for doing so.

PR #158, opened 2016-12-01
PeakInvestigator: new classes for extracting mass lists from tarball

TDI-tenderholt:
This class will be used for processing results from PeakInvestigator.


TDI-tenderholt:
Looks like there's problems with the EBI repo again.

msdk-rawdata/msdk-rawdata-peakinvestigator/src/main/java/io/github/msdk/rawdata/peakinvestigator/PeakInvestigatorMsSpectrum.java
```diff
@@ -0,0 +1,125 @@
...
+import io.github.msdk.datamodel.msspectra.MsSpectrum;
+import io.github.msdk.datamodel.msspectra.MsSpectrumType;
+import io.github.msdk.util.tolerances.MzTolerance;
+
+/**
+ * This class is used provide m/z-dependent mass tolerances from
+ * PeakInvestigator by decorating an existing instance of MsSpectrum.
+ *
+ */
+public class PeakInvestigatorMsSpectrum implements MsSpectrum, MzTolerance {
```
tomas-pluskal: This is fine, but keep in mind that if you add a new implementation of MsSpectrum, it has to be updated every time there is a change in the MsSpectrum interface. If all modules add their own implementations, it becomes a hassle to maintain them all.
I think you could achieve the same without adding a new implementation, simply using a container class that contains both an MsSpetrum and an MzTolerance, accessible by getters getMsSpectrum() and getMzTolerance().

msdk-rawdata/msdk-rawdata-peakinvestigator/src/main/java/io/github/msdk/rawdata/peakinvestigator/PeakInvestigatorMsSpectrum.java
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	@Override
+	public MsSpectrumType getSpectrumType() {
+		return MsSpectrumType.CENTROIDED;
+	}
+
+	@Override
+	public void setSpectrumType(MsSpectrumType spectrumType) {
+		throw new RuntimeException("This class should not have a MsSpectrumType set.");
```
TDI-tenderholt: Great point, I'll make that change.

msdk-rawdata/msdk-rawdata-peakinvestigator/src/main/java/io/github/msdk/rawdata/peakinvestigator/PeakInvestigatorMsSpectrum.java
TDI-tenderholt: Also a good point.

I was hoping to map the PI workflow to standard centroiding as closely as possible (i.e. the resulting ```RawDataFile``` can be transparently passed to other Methods like ```ChromatogramBuilder```), but still have the mass errors around if desired. So while a custom ```MsSpectrum``` probably isn't necessary, I think a custom ```RawDataFile``` that either implements ```MzToleranceProvider``` or has a getter for one will be necessary.

What do you think?

TDI-tenderholt: What sort of filters are you thinking would be appropriate after PI? SGFilter, MeanFilter, and Resample don't seem appropriate for centroided data. CropFilter seems more appropriate prior to PI (i.e. don't process more scans than necessary) or during PI (it can limit analysis to specific mass range).

I do agree that it makes sense to remove ```MzToleranceProvider``` and add ```MsSpectrum.getMzTolerance()```. How do you want to handle ```null``` in methods that expect a MzTolerance? Fallback onto a MzTolerance specified by an individual MSDKMethod like below?

```
        for (MsScan scan : inputScans) {

            if (canceled)
                return null;

            MzTolerance mzTolerance = mzToleranceProvider.getMzTolerance(scan);
            if (mzTolerance == null) {
                 mzTolerance = this.mzTolerance;
            }

            massConnector.addScan(inputFile, scan, mzTolerance);
            processedScans++;
        }
```

tomas-pluskal: For example, in MZmine we use a shoulder peaks filter for FTMS data, to remove residual signals from the Fourier Transform. It's not implemented in MSDK yet, just a stub (https://github.com/msdk/msdk/tree/master/msdk-spectra/msdk-spectra-shoulderpeaks).

Yes, the Chromatogram Builder can have a default mzTolerance in case the spectrum has null.

TDI-tenderholt:
I've updated MsSpectrum and removed MzToleranceProvider, and associated changes to make it compile locally—Travis CI is still having trouble with something in an EBI repo. Let me know if those changes are ok.

I'll work on the remaining changes to the PI-specific code next week.

tomas-pluskal:
Thanks a lot. I reported the EBI repo problem a while ago (https://github.com/PRIDE-Utilities/jmzml/issues/19), but so far they have not fixed it.

PR #160, opened 2017-01-12
Fix merging isotopes close in mass

chhh:
`mergeIsotopes()` method produced strange results. When called with `mzTolerance` 0.1, for example, it would leave separate peaks that were 0.004 Da apart.

Test: `formula: C39H60N14O14`, `mz tolerance: 0.1`  

```
Old result:
mz:  948.441	 949.444	 950.444	 950.448	 951.450	
ab:    1.000	   0.473	   0.050	   0.087	   0.024

New result:
mz:  948.441	 949.444	 950.447	 951.450	
ab:    1.000	   0.473	   0.137	   0.024
```

Also had to add a hack in the end of the method to fix up remaining mass values, otherwise SimpleMsSpectrum would complain about masses not being in sorted order.

### DIFF
--- a/msdk-spectra/msdk-spectra-isotopepattern/src/main/java/io/github/msdk/spectra/isotopepattern/IsotopePatternGeneratorAlgorithm.java
+++ b/msdk-spectra/msdk-spectra-isotopepattern/src/main/java/io/github/msdk/spectra/isotopepattern/IsotopePatternGeneratorAlgorithm.java
@@ -151,32 +151,43 @@ public class IsotopePatternGeneratorAlgorithm {
      * intensity is added together and new m/z value is calculated as a weighted
      * average.
      */
-    private static int mergeIsotopes(double mzValues[], float intensityValues[],
-            int size, double mzTolerance) {
+    private static int mergeIsotopes(double mzValues[], float intensityValues[], int size, double mzTolerance) {
+        if (mzValues.length != intensityValues.length)
+            throw new IllegalArgumentException("Mass and intensity arrays must be of the same size");
+        if (mzValues.length < 2)
+            return mzValues.length;
 
-        int newSize = 0;
+        int ptrCur = 0;
+        int ptrNex = 1;
+        for ( ; ptrNex < mzValues.length; ptrNex++) {
+            double mzCur = mzValues[ptrCur];
+            double mzNex = mzValues[ptrNex];
+            if (Math.abs(mzCur - mzNex) < mzTolerance) {
+                // merge, only next pointer moves
+                float abCur = intensityValues[ptrCur];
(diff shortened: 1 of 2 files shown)


msdk-spectra/msdk-spectra-isotopepattern/src/main/java/io/github/msdk/spectra/isotopepattern/IsotopePatternGeneratorAlgorithm.java
```diff
@@ -151,32 +151,43 @@
...
+                ptrCur++;
+
+                mzValues[ptrCur] = mzValues[ptrNex];
+                intensityValues[ptrCur] = intensityValues[ptrNex];
             }
-            newSize++;
         }
-        return newSize;
+
+        // ugly hack to make SimpleSpectrum not complain about masses not being sorted
```
tomas-pluskal: You can use DataPointSorter.sortDataPoints() here

tomas-pluskal:
Thanks for the fix. How about adding a junit test testing this function, to make sure this bug does not reappear in the future? I think that is generally a good policy.

msdk-spectra/msdk-spectra-isotopepattern/src/main/java/io/github/msdk/spectra/isotopepattern/IsotopePatternGeneratorAlgorithm.java
chhh: I don't think so , the masses get messed up, e.g. they can become 100.0, 101.0, 102.0, 101.91, 101.92. Everything that is beyond the current pointer is garbage, and you can't re-sort the data. The problem is only in `SimpleMsSpectrum` checking for correct m/z ordering in the constructor, however it does allow for limiting the mass array's actual size. I guess the correct way to fix it is to limit the length up to which `SimpleMsSpectrum` checks the ordering.

tomas-pluskal: Since arrays are not easily resized in Java, we always use the trio of mzValues[], intensityValues[], and int size. The size indicates the real size of the array, while the length of the actual arrays might be different. Note that the SimpleMsSpectrum only checks the proper ordering of m/z values up to the given size, not the whole array:
https://github.com/msdk/msdk/blob/master/msdk-datamodel/src/main/java/io/github/msdk/datamodel/impl/SimpleMsSpectrum.java#L113
Also the DataPointSorter.sortDataPoints() sorts everything up to given size and ignores the rest.

PR #161, opened 2017-01-12
Isotope pattern

chhh:
Removed ugly hack, added test, stopped ignoring another isotopic pattern test which works fine.

### DIFF
--- a/msdk-spectra/msdk-spectra-isotopepattern/src/main/java/io/github/msdk/spectra/isotopepattern/IsotopePatternGeneratorAlgorithm.java
+++ b/msdk-spectra/msdk-spectra-isotopepattern/src/main/java/io/github/msdk/spectra/isotopepattern/IsotopePatternGeneratorAlgorithm.java
@@ -151,32 +151,43 @@ public class IsotopePatternGeneratorAlgorithm {
      * intensity is added together and new m/z value is calculated as a weighted
      * average.
      */
-    private static int mergeIsotopes(double mzValues[], float intensityValues[],
-            int size, double mzTolerance) {
+    private static int mergeIsotopes(double mzValues[], float intensityValues[], int size, double mzTolerance) {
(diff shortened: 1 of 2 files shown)


tomas-pluskal:
Hi, I realized this whole pull request is a result of misunderstanding. 
As I explained in my last comment, we use the 'size' variable to indicate the size of the spectrum, no matter what the length of the m/z and intensity arrays is.
For example 
````
mzValues.length == 500
intensityValues.length == 300
size == 200
````
means the spectrum has 200 data points, which must be sorted in m/z order. Everything above 200 is ignored.  The purpose of this is to be able to reuse existing arrays, because primitive arrays cannot be resized in Java. I think reusing the arrays reduced the pressure on garbage collector when lots of spectra are processed.

**Comment by tomas-pluskal** [comment]
Note: I added your test (which passes fine with current code) and reactivated the ignored test:
https://github.com/msdk/msdk/commit/bea6d9a441cbeb36858c91545d216de3910e2552

**Comment by chhh** [comment]
Yeah, I realized that the hack was not needed anymore, this was a leftover from previous stuff.
The test is not passing, it's missing @Test annotation.

For me the test is failing:
This test:
```
@Test
    public void testC39H60N14O14() throws MSDKException {
        String formula = "C39H60N14O14";

        MsSpectrum pattern = IsotopePatternGeneratorAlgorithm.generateIsotopes(formula, 0.01, 1.0f, 0.1);

        Assert.assertEquals(new Integer(4), pattern.getNumberOfDataPoints());
        Assert.assertEquals(1.0f, pattern.getIntensityValues()[0], 0.000001);
        Assert.assertEquals(951.4501830990874, pattern.getMzValues()[3], 0.00001);
    }
```

Fails with:
```
java.lang.AssertionError: 
Expected :4
Actual   :5
```

tomas-pluskal:
You're right. OK, I'm merging your PR..

PR #163, opened 2017-03-21
Attempt to add FeatureFindMetabo

adhi1419:


adhi1419:
The build fails at "msdk-spectra\msdk-spectra-spectrumtypedetection\target\surefire-reports" which I haven't touched

tomas-pluskal:
Hi Adhithya,
This is a great start! I'll add a few comments to the code.

I am aware that a few tests fail in the spectrumtypedetection module, that's because the spectrum type detection algorithm is actually far from perfect. I will probably just disable those tests for the time being.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboMZValueDetector.java
```diff
@@ -0,0 +1,83 @@
+package io.github.msdk.featdet.openms;
```
tomas-pluskal: Please add the license header to all new files (the same header that is found in all other MSDK files)

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboMZValueDetector.java
```diff
@@ -0,0 +1,83 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.NodeList;
+import org.w3c.dom.Node;
+import org.w3c.dom.Element;
+import org.xml.sax.SAXException;
+
+import io.github.msdk.datamodel.rawdata.RawDataFile;
+
+public class FeatureFinderMetaboMZValueDetector {
+	private final String OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME = "FeatureFinderMetabo";
```
tomas-pluskal: This is ok, but it assumes that the FeatureFinderMetabo program is in PATH. Often, that is not the case. It would be nice to add a bit more intelligence to this. Perhaps you can create a new class FeatureFinderMetaboLocator for that purpose. The class can have a single function findFeatureFinderMetabo(), which will contain the logic. The logic can then search some common paths based on platform (/Applications/OpenMS*/TOPP on Mac, C:\Program Files\... on Windows, /usr/local/openMS*/... on Linux). Of course we cannot know for sure where OpenMS is installed, but we try as many possibilities as we can.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboMZValueDetector.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.w3c.dom.Document;
+import org.w3c.dom.NodeList;
+import org.w3c.dom.Node;
+import org.w3c.dom.Element;
+import org.xml.sax.SAXException;
+
+import io.github.msdk.datamodel.rawdata.RawDataFile;
+
+public class FeatureFinderMetaboMZValueDetector {
```
tomas-pluskal: I'd call the class simply FeatureFinderMetaboDetector, as "MZValue" can be a bit confusing (I can explain why if you're interested)

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboMZValueDetector.java
```diff
@@ -0,0 +1,83 @@
...
+
+	public FeatureFinderMetaboMZValueDetector(File mzMLFile, File featureXMLFile) {
+		this.mzMLFile = mzMLFile;
+		this.featureXMLFile = featureXMLFile;
+	}
+
+	public List<Double> execute() throws SAXException, IOException, ParserConfigurationException {
+		List<Double> mzValuesResult = new ArrayList<Double>();
+
+		System.out.println(execShellCommand(OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME + " -in "
```
tomas-pluskal: System.out.println() is not nice, in MSDK we use logging via the SLF4J library.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboMZValueDetector.java
```diff
@@ -0,0 +1,83 @@
...
+			if (node.getNodeName() == "featureList")
+				featureList = node.getChildNodes();
+		}
+
+		for (int i = 0; i < featureList.getLength(); i++) {
+			Node node = featureList.item(i);
+			if (node.getNodeName() == "feature") {
+				Element featureElement = (Element) node;
+				mzValuesResult
+						.add(Double.valueOf(featureElement.getElementsByTagName("position").item(1).getTextContent()));
```
tomas-pluskal: Using this kind of nested calls is asking for trouble, because a NullPointerException can be thrown at any point. There should at least be a try/catch mechanism in case that happens.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboMZValueDetector.java
```diff
@@ -0,0 +1,83 @@
...
+		NodeList featureList = null;
+		for (int i = 0; i < featureMap.getLength(); i++) {
+			Node node = featureMap.item(i);
+			if (node.getNodeName() == "featureList")
+				featureList = node.getChildNodes();
+		}
+
+		for (int i = 0; i < featureList.getLength(); i++) {
+			Node node = featureList.item(i);
+			if (node.getNodeName() == "feature") {
```
tomas-pluskal: In Java you should generally use .equals for Strings instead of the == operator
http://stackoverflow.com/questions/513832/how-do-i-compare-strings-in-java

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboMZValueDetector.java
```diff
@@ -0,0 +1,83 @@
...
+
+import io.github.msdk.datamodel.rawdata.RawDataFile;
+
+public class FeatureFinderMetaboMZValueDetector {
+	private final String OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME = "FeatureFinderMetabo";
+
+	private final @Nonnull File mzMLFile;
+	private final @Nonnull File featureXMLFile;
+
+	public FeatureFinderMetaboMZValueDetector(String mzMLFileName, String featureXMLFileName) {
```
tomas-pluskal: In fact, since we are in MSDK, we don't really need the output featureXMLFileName. It is better to redirect the OpenMS output into a temporary file, then parse the contents, and delete the temporary file afterwards. So you can remove the featureXMLFileName parameter.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboMZValueDetector.java
```diff
@@ -0,0 +1,83 @@
...
+		String out = "";
+		try {
+			Process cmdProc = Runtime.getRuntime().exec(cmd);
+			BufferedReader stdoutReader = new BufferedReader(new InputStreamReader(cmdProc.getInputStream()));
+			String line;
+			while ((line = stdoutReader.readLine()) != null) {
+				out += line + "\n";
+			}
+		} catch (IOException ie) {
+			ie.printStackTrace();
```
tomas-pluskal: What I am missing here is a better error recovery. What if the FeatureFinderMetabo finishes with an error? (e.g., malformed mzML file?). The printStackTrace() will not do anything and the function simply returns. Errors should be properly covered for, and I also recommend to add a junit test for a malformed input.

adhi1419:
Thank you for the detailed feedback Tomas, I will work on it and keep you posted.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,92 @@
...
+		String out = "";
+		try {
+			Process cmdProc = Runtime.getRuntime().exec(cmd);
+			BufferedReader stdoutReader = new BufferedReader(new InputStreamReader(cmdProc.getInputStream()));
+			String line;
+			while ((line = stdoutReader.readLine()) != null) {
+				out += line + "\n";
+			}
+		} catch (IOException ie) {
+			throw new FeatureFinderMetaboException(ie.getMessage());
```
adhi1419: This throws an exception even if one of the attempts to locate FeatureFinderMetabo fails.
Have to find a method to execute the binary from path without throwing an exception (The code would anyways proceed past the Exception when not run as a test case, and determine the correct location of the library if it exists).

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboException.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * or (per the licensee's choosing)
+ *
+ * (b) the terms of the Eclipse Public License v1.0 as published by
+ * the Eclipse Foundation.
+ */
+
+package io.github.msdk.featdet.openms;
+
+public class FeatureFinderMetaboException extends Exception {
```
tomas-pluskal: We have an MSDKException class, so I think you don't need to create an extra exception class here (unless there is some extra functionality)

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,92 @@
...
+				}
+			} else {
+				return null;
+			}
+		}
+		return null;
+	}
+
+	private static boolean isFeatureFinderMetaboHere(String path) throws FeatureFinderMetaboException {
+		return execShellCommand(path).contains(FEATURE_FINDER_METABO_EVAL);
```
tomas-pluskal: Perhaps it is not necessary to execute the program to find if it is present. How about just checking for the executable file using java.io.File.canExecute()?

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,92 @@
...
+		} else {
+			String OS = System.getProperty("os.name").toLowerCase();
+			if (OS.indexOf("win") >= 0) {
+				for (String s : new File(OPENMS_DEFAULT_LOCATION_WINDOWS).list()) {
+					if (s.toLowerCase().contains("openms")) {
+						String path = OPENMS_DEFAULT_LOCATION_WINDOWS + s + "\\bin\\"
+								+ OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME;
+						if (isFeatureFinderMetaboHere(path))
+							return path;
+						return null;
```
tomas-pluskal: The position of this return is suspicious (it will return null even if there are other openms folder that have not been checked yet). The same applies to the other platforms below.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboDetector.java
```diff
@@ -0,0 +1,108 @@
...
+
+	private final @Nonnull File mzMLFile;
+
+	public FeatureFinderMetaboDetector(String mzMLFileName) throws FeatureFinderMetaboException {
+		this(new File(mzMLFileName));
+	}
+
+	public FeatureFinderMetaboDetector(File mzMLFile) throws FeatureFinderMetaboException {
+		this.mzMLFile = mzMLFile;
+		FEATURE_XML_FILE = new File(mzMLFile.getName() + "featureXML");
```
tomas-pluskal: How about using File.createTempFile() here?
You can also combine that with File.deleteOnExit()

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboDetector.java
```diff
@@ -0,0 +1,108 @@
...
+		OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME = FeatureFinderMetaboLocator.findFeatureFinderMetabo();
+	}
+
+	public List<Double> execute()
+			throws SAXException, IOException, ParserConfigurationException, FeatureFinderMetaboException {
+		List<Double> mzValuesResult = new ArrayList<Double>();
+
+		if (!OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME.equals(null)) {
```
tomas-pluskal: Instead of checking !OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME, you can check for OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME and fail quickly. This way you avoid long nested _if_ clauses.

tomas-pluskal: Note: this is just about code readability, not about function

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboDetector.java
adhi1419: I could have used File.createTempFile(), but the featureXML file is being created by the FeatureFinderMetabo library. I'm just creating a Java File Reference of the existing file here.

tomas-pluskal: You're right. But we should still make sure the temporary file name is unique, and preferably created in the folder for temporary file (java.io.tmpdir). See e.g. here for inspiration: 
http://stackoverflow.com/questions/1293655/how-to-create-tmp-file-name-with-out-creating-file

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
adhi1419: I couldn't get this to work. Strangely, the canExecute() method returns false even when I point it to the exact location.
Apart from that, this method still can't detect if the method is in the PATH (Please correct me if I am wrong).
I have managed to fix it for now, by not throwing the Exception, but instead just catching it. I will change to a better logic as soon as I find one.

tomas-pluskal:
Hi Adhithya,

It looks much better now. I will add a few more comments - some of them are only minor, but good for you to learn some useful long-term habits.

msdk-featdet/pom.xml
```diff
@@ -15,5 +15,6 @@
         <module>msdk-featdet-gridmass</module>
```
tomas-pluskal: Please also add a dependency to the msdk-all module, which aggregates all MSDK modules.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,90 @@
...
+package io.github.msdk.featdet.openms;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+import io.github.msdk.MSDKException;
+
+public class FeatureFinderMetaboLocator {
```
tomas-pluskal: Please try to add JavaDoc comments to all new classes etc.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,90 @@
...
+
+	private final static String OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME = "FeatureFinderMetabo";
+	private final static String OPENMS_DEFAULT_LOCATION_WINDOWS = "C:\\Program Files\\";
+	private final static String OPENMS_DEFAULT_LOCATION_MAC = "/Applications/";
+	private final static String OPENMS_DEFAULT_LOCATION_LINUX = "/usr/local/";
+	private final static String FEATURE_FINDER_METABO_EVAL = "No options given. Aborting!";
+
+	public static String findFeatureFinderMetabo() throws MSDKException {
+		if (!isFeatureFinderMetaboHere(OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME)) {
```
tomas-pluskal: This line should have a comment to explain what you are doing (checking if FeatureFinderMetabo is in PATH?). And I think the logic is inverted - if isFeatureFinderMetabo returns true, you should just return the default location and exit.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,90 @@
...
+			return OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME;
+		} else {
```
tomas-pluskal: There is no need for else clause, because the if clause ends with a return anyway. Avoiding the else clause improves the code readability.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,90 @@
...
+			String OS = System.getProperty("os.name").toLowerCase();
```
tomas-pluskal: Although not necessary, for variables that are not going to change, it is a good practice to use the _final_ keyword (final String OS = ...)
http://stackoverflow.com/questions/137868/using-final-modifier-whenever-applicable-in-java

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboDetector.java
```diff
@@ -0,0 +1,116 @@
...
+			}
+
+		}
+		return mzValuesResult;
+
+	}
+
+	private void createTempFeatureXMLFile() throws IOException, MSDKException {
+		final Logger logger = LoggerFactory.getLogger(this.getClass());
+		logger.info(execShellCommand(OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME + " -in " + mzMLFile.getAbsolutePath()
```
tomas-pluskal: It is a good practice in Java to avoid combining many function calls on one line - if an exception is thrown (e.g., a NullPointerException or IOException), the Java runtime will tell you the line number, but you won't know if the exception was thrown by logger.info() or by execShellCommand() or by mzMLFile.getAbsolutePath(). It is much better to split this into separate code lines, each line performing one function. Again, it also improves code readability.

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboDetector.java
```diff
@@ -0,0 +1,116 @@
...
+				+ " -out " + FEATURE_XML_FILE.getAbsolutePath()));
+		if (!FEATURE_XML_FILE.exists())
+			throw new MSDKException("Could not create featureXML File");
```
tomas-pluskal: Perhaps "could not create file" is not the best error message here. The code should distinguish several possibilities:
- FeatureFinderMetabo was not found
- FeatureFinderMetabo was executed, but could not process the input mzML file (can we parse some error message from FeatureFinderMetabo?)
- FeatureFinderMetabo was executed and processed the input mzML file, but we failed to parse the resulting featureXML file

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,90 @@
...
+				for (String s : new File(OPENMS_DEFAULT_LOCATION_LINUX).list()) {
+					if (s.toLowerCase().contains("openms")) {
+						String path = "\"" + OPENMS_DEFAULT_LOCATION_LINUX + s + "\\bin\\"
+								+ OPENMS_FEATURE_FINDER_METABO_LIBRARY_NAME + "\"";
+						if (isFeatureFinderMetaboHere(path))
+							return path;
+					}
+
+				}
+			} else {
```
tomas-pluskal: this else block is unnecessary

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
```diff
@@ -0,0 +1,90 @@
...
+				return null;
+			}
+		}
+		return null;
+	}
+
+	private static boolean isFeatureFinderMetaboHere(String path) throws MSDKException {
+		return execShellCommand(path).contains(FEATURE_FINDER_METABO_EVAL);
```
tomas-pluskal: A comment would help to understand what you are doing here

msdk-featdet/msdk-featdet-openms/src/main/java/io/github/msdk/featdet/openms/FeatureFinderMetaboLocator.java
adhi1419: Oops, I had changed it while testing it and forgot to change it back.

tomas-pluskal:
Thanks, Adhithya,

I'll merge this now!

PR #168, opened 2017-05-30
Parsing of metadata

adhi1419:


tomas-pluskal:
A few comments: 
* MzMLSpectrum should implement MsScan
* the execute() method should return MzMLRawDataFile (not InputStream)
* MzMLRawDataFile is a class that implements RawDataFile, you can copy it from the existing mzML reader and start from there

PR #175, opened 2017-06-01
Save position, type & bit length of binary arrays

adhi1419:


msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLBinaryDataInfo.java
```diff
@@ -0,0 +1,65 @@
+package io.github.msdk.io.mzml2;
+
+public class MzMLBinaryDataInfo {
```
tomas-pluskal: These classes can be package-scope only, because they will be only used by the parser

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLBinaryDataInfo.java
```diff
@@ -0,0 +1,65 @@
+package io.github.msdk.io.mzml2;
```
tomas-pluskal: Pls add the license header

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLCV.java
```diff
@@ -0,0 +1,125 @@
...
+ * @see <a href=
+ *      "http://psidev.cvs.sourceforge.net/viewvc/psidev/psi/psi-ms/mzML/controlledVocabulary/psi-ms.obo">
+ *      Official CV specification</a>
+ * 
+ */
+class MzMLCV {
+
+	static enum MzMLCompressionType {
+		ZLIB("MS:1000574"); // UNCOMPRESSED, MSNUMPRESS
+							// Unable to find corresponding Accession Codes
```
tomas-pluskal: Check the official controlled vocabulary documentation: https://raw.githubusercontent.com/HUPO-PSI/psi-ms-CV/master/psi-ms.obo
It seems there are 6 different types of numpress compression CV terms

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLCV.java
```diff
@@ -0,0 +1,125 @@
...
+	// Scan filter string
+	static final String cvScanFilterString = "MS:1000512";
+
+	// Precursor m/z.
+	static final String cvPrecursorMz = "MS:1000744";
+
+	// Polarity
+	static final String cvPolarityPositive = "MS:1000130";
+	static final String cvPolarityNegative = "MS:1000129";
+	static final CVParam polarityPositiveCvParam = new CVParam();
```
tomas-pluskal: The CVParam class belongs to jmzml, we're not going to use that

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
```diff
@@ -32,11 +32,11 @@
 
 public class MzMLSpectrum implements MsScan {
 	private HashMap<String, String> cvParamValues;
-	private ArrayList<Long> binaryDataPositions;
+	private ArrayList<MzMLBinaryDataInfo> binaryDataInfoArray;
```
tomas-pluskal: I think you don't need ArrayLists here - each scan has m/z array and intensity array, and that's it. There will always be two arrays.

adhi1419: There is also an array of type "wavelength array", what do I do with it?

adhi1419: But these scans do not have m/z array, so do we not store these scans at all?

tomas-pluskal: We don't, just skip them

PR #176, opened 2017-06-07
Implementing code from MSFTBX for decoding binary data

adhi1419:


msdk-all/pom.xml
```diff
@@ -274,5 +274,10 @@
 			<version>${project.version}</version>
 		</dependency>
 
+		<dependency>
```
tomas-pluskal: This dependency should not go to msdk-all, only to your module

PR #180, opened 2017-06-09
Use some Java 8 features in datamodel

ATenderholt:
* Use ```Optional``` instead of relying on ```null``` if values are absent.
* Use default methods


tomas-pluskal:
Looks good, thanks!

PR #181, opened 2017-06-09
Decoding binary data on request

adhi1419:
The decode method gives a java.util.zip.DataFormatException which I don't quite understand why. The parameters passed to the method might be wrong, can you have a look into it

### DIFF
--- a/msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
+++ b/msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
@@ -14,11 +14,22 @@
 
 package io.github.msdk.io.mzml2;
 
+import java.io.IOException;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Scanner;
+import java.util.zip.DataFormatException;
+
+import javax.xml.stream.XMLEventReader;
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.events.StartElement;
+import javax.xml.stream.events.XMLEvent;
 
 import com.google.common.collect.Range;
 
+import io.github.msdk.MSDKException;
 import io.github.msdk.datamodel.msspectra.MsSpectrumType;
 import io.github.msdk.datamodel.rawdata.ActivationInfo;
 import io.github.msdk.datamodel.rawdata.IsolationInfo;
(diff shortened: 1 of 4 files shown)


msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
```diff
@@ -90,8 +101,50 @@ public Integer getNumberOfDataPoints() {
...
+		byte[] bytesIn = new byte[getMzBinaryDataInfo().getEncodedLength()];
+		int precision = 0;
+		EnumSet<MzMLBinaryDataInfo.MzMLCompressionType> compressions = EnumSet
+				.noneOf(MzMLBinaryDataInfo.MzMLCompressionType.class);
+		try {
+			mappedByteBufferInputStream.setPosition(getMzBinaryDataInfo().getPosition());
+			XMLInputFactory xmlInputFactory = XMLInputFactory.newInstance();
+			XMLEventReader xmlEventReader = xmlInputFactory.createXMLEventReader(mappedByteBufferInputStream);
+
+			if (getMzBinaryDataInfo().getBitLength().toString().startsWith("THIRTY_TWO")) {
```
tomas-pluskal: There is no need to convert to String here, is it?

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
```diff
@@ -90,8 +101,50 @@ public Integer getNumberOfDataPoints() {
...
+						xmlEvent = xmlEventReader.nextEvent();
+						bytesIn = xmlEvent.asCharacters().getData().getBytes();
+						compressions.add(getMzBinaryDataInfo().getCompressionType());
+						result = MzMLPeaksDecoder.decode(bytesIn, getMzBinaryDataInfo().getArrayLength(),
+								new Integer(precision), getMzBinaryDataInfo().getEncodedLength(), compressions).arr;
+						break;
+					}
+				}
+			}
+		} catch (XMLStreamException e) {
```
tomas-pluskal: Instead of many catch blocks, use catch (Exception e) to catch all of them

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
```diff
@@ -90,8 +101,50 @@ public Integer getNumberOfDataPoints() {
 
 	@Override
 	public double[] getMzValues() {
```
tomas-pluskal: The getMzValues() method should not be parsing any XML - it should only obtain the Base64-encoded data from the correct position in the memory-mapped mzML file, and decode it.

adhi1419: How do we know where to stop the data from the InputStream without the parser?

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
adhi1419: It returns the ENUM Constant, and we have to know whether it is a constant referring to a 32 bit or 64 bit using its declared name (Hence "toString()")

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
adhi1419: I will keep that in mind, thanks

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
tomas-pluskal: Just the == operator to compare it to the enum value

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
tomas-pluskal: As I said before, you should save the position and size of the encoded block. Then you can just read the number of bytes that is necessary.

PR #182, opened 2017-06-13
Code optimization and decoding Intensity values

adhi1419:


msdk-io/msdk-io-mzml2/src/test/java/io/github/msdk/io/mzml2/MzMLFileParserTest.java
```diff
@@ -26,20 +25,20 @@
...
 
-		final String inputFileName = TEST_DATA_PATH + "mzML_with_UV.mzML";
-		final File inputFile = new File(inputFileName);
+    final String inputFileName = TEST_DATA_PATH + "mzML_with_UV.mzML";
+    final File inputFile = new File(inputFileName);
 
-		Assert.assertNotNull(inputFile);
-		MzMLFileParser mzParser = new MzMLFileParser(inputFile);
-		mzParser.execute();
+    Assert.assertNotNull(inputFile);
```
tomas-pluskal: What is the reason you check if inputFile == null? Is that ever going to happen?

msdk-io/msdk-io-mzml2/src/test/java/io/github/msdk/io/mzml2/MzMLFileParserTest.java
```diff
@@ -26,20 +25,20 @@
...
+    MzMLFileParser mzParser = new MzMLFileParser(inputFile);
+    mzParser.execute();
 
-		Assert.assertNotNull(new File(inputFileName));
-		MzMLFileParser mzParser2 = new MzMLFileParser(inputFileName);
-		mzParser2.execute();
-	}
+    Assert.assertNotNull(new File(inputFileName));
+    MzMLFileParser mzParser2 = new MzMLFileParser(inputFileName);
+    mzParser2.execute();
```
tomas-pluskal: If the decoding code already works, you should write a test for it

adhi1419: Should we test for decoded values of all scans?

tomas-pluskal:
Not necessary for all scans, just select a few to make sure it works correctly.

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MappedByteBufferInputStream.java
```diff
@@ -19,35 +18,41 @@
...
+      return -1;
+    }
+
+    len = Math.min(len, buf.remaining());
+    buf.get(bytes, off, len);
+    return len;
+  }
+
+  public synchronized InputStream getInputStream(int position, int length) {
+    MappedByteBuffer newBuf = (MappedByteBuffer) buf.position(position);
```
tomas-pluskal: You need to make a duplicate of the byte buffer before setting the position. Otherwise, if two threads are using the same MappedByteBuffer, it will mess up their input (the positions have to be independent).

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MappedByteBufferInputStream.java
```diff
@@ -19,35 +18,42 @@
...
+      return -1;
+    }
+
+    len = Math.min(len, buf.remaining());
+    buf.get(bytes, off, len);
+    return len;
+  }
+
+  public synchronized InputStream getInputStream(int position, int length) {
+    MappedByteBuffer newBuf = buf;
```
tomas-pluskal: This code does not duplicate the buffer - you only create a new reference to the same object.
Use this method: https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#duplicate()

PR #197, opened 2017-06-21
Parsing Chromatograms - Completing the parser

adhi1419:


msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
```diff
@@ -181,7 +208,12 @@ public Integer getNumberOfDataPoints() {
   @Override
   public double[] getMzValues() {
     double[] mzValues = null;
-    this.mzArrayLength = getMzBinaryDataInfo().getArrayLength();
+    Integer numOfDataPoints = this.numOfDataPoints;
```
tomas-pluskal: I think there's no need to create Integer numOfDataPoints, since you can just use this.numOfDataPoints

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
```diff
@@ -219,13 +252,16 @@ public Integer getNumberOfDataPoints() {
   @Override
   public float[] getIntensityValues() {
     float[] intensityValues = null;
-    this.mzArrayLength = getIntensityBinaryDataInfo().getArrayLength();
+    Integer numOfDataPoints = this.numOfDataPoints;
```
tomas-pluskal: Same here

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLFileParser.java
```diff
@@ -148,27 +155,32 @@ public MzMLRawDataFile execute() throws MSDKException {
...
             return null;
 
           eventType = xmlStreamReader.next();
 
+          progress = ((float) xmlStreamReader.getLocation().getCharacterOffset() / is.length());
+
+          // Log progress after every 10% completion
+          if ((int) (progress * 100) >= approxProgress + 10) {
+            approxProgress = (int) (progress * 10) * 10;
+            logger.info("Parsing in progress... " + approxProgress + "% completed");
```
tomas-pluskal: perhaps this can go to logger.debug()

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLFileParser.java
```diff
@@ -62,12 +62,12 @@
  */
 public class MzMLFileParser implements MSDKMethod<RawDataFile> {
   private final @Nonnull File mzMLFile;
-  private final @Nonnull ArrayList<MsScan> spectrumList;
   private final ArrayList<MzMLReferenceableParamGroup> referenceableParamGroupList;
   private MzMLRawDataFile newRawFile;
-  private Integer lastScanNumber = 0;
   private volatile boolean canceled;
   private Float progress;
+  private int approxProgress;
```
tomas-pluskal: since this variable is only used for logger output, the name "approxProgress" is not very clear. how about something like "lastLoggedProgress"?

msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/MzMLSpectrum.java
adhi1419: Sorry, I'll fix it. I was initially intending to set that value to the one contained in BinaryDataInfo. Should we also throw an exception in this case? http://www.peptideatlas.org/tmp/mzML1.1.0.html#binaryDataArray

tomas-pluskal: You should log a warning, but don't throw an exception as long as the data is parseable.

PR #198, opened 2017-06-21
MzMLFileMemoryMapper: make class abstract and one method static

TDI-tenderholt:
No state is stored, so no reason to actually allocate instances to use this class.

Also minor cleanup by removing ```XMLStreamException``` and ```MSDKException``` since they aren't actually thrown by anything here.


tomas-pluskal:
Isn't this going to conflict with PR #197?

TDI-tenderholt:
Do you mean code-wise? Or because you actually use an instance of ```MzMLFileMemoryMapper``` somewhere else?

Searching the diff here on Github only shows it being used in the one spot in ```MzMLFileParser```.

tomas-pluskal:
I just meant that merging this PR might create a conflict for PR197, because it modifies the same files. But I guess Adhithiya can resolve that tomorrow, so I'll merge this one.

PR #206, opened 2017-06-25
Fix decoding uncompressed binary arrays

adhi1419:


msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml2/util/MzMLPeaksDecoder.java
```diff
@@ -80,48 +78,52 @@
...
-      case NUMPRESS_POSINT_ZLIB:
-      case NUMPRESS_SHLOGF_ZLIB:
-        iis = new InflaterInputStream(is);
-        dis = new LittleEndianDataInputStream(iis);
-        break;
-      default:
-        dis = new LittleEndianDataInputStream(is);
-        break;
-    }
+    if (binaryDataInfo.getCompressionType() != null) {
```
tomas-pluskal: Since you have MzMLCompressionType.NO_COMPRESSION, the binaryDataInfo.getCompressionType() should not return null. Perhaps you could declare it as @Nonnull and make sure the default value is always MzMLCompressionType.NO_COMPRESSION

PR #207, opened 2017-06-25
Additional mzXML parser using javolution's XMLStreamReader

adhi1419:
This parser is slower than the existing one, needs a review.


msdk-io/msdk-io-mzxml/src/main/java/io/github/msdk/io/mzxml/MzXMLFileParser.java
```diff
@@ -0,0 +1,318 @@
...
+          }
+
+        } while (eventType != XMLStreamConstants.END_DOCUMENT);
+
+      } finally {
+        if (xmlStreamReader != null) {
+          xmlStreamReader.close();
+        }
+      }
+    } catch (IOException | XMLStreamException | javax.xml.stream.XMLStreamException
```
tomas-pluskal: In case like this you can just do 
````
} catch(Exception e) {
  throw (new MSDKException(e));
}
````
or even better:
````
} catch(Throwable e) {
  throw (new MSDKException(e));
}
````
That will also take care of runtime errors like OutOfMemoryError

tomas-pluskal:
Hi @adhi1419 , 
If you tested the speed in comparison to the existing parser, can you also add the testing (benchmarking) code to the tests folder? It is not a junit test, so don't mark it as @Test. You can add it into a separate class with its own main() method.

PR #218, opened 2017-07-01
Get mz values for Chromatograms

adhi1419:


TDI-tenderholt:
I don't know how @tomas-pluskal and @chhh feel about messy commit histories, but I think you should rebase your branch onto current master so that there is just a single commit in this pull request.

Let us know if you need help with this.

PR #224, opened 2017-07-08
Bi-Gaussian class is added for fitting it on EIC.

dharak029:


msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+			
+			@Override
+			public int compare(Double o1, Double o2) {
+				Double intensity1 = o1;
+				Double intensity2 = o2;
+				return intensity1.compareTo(intensity2);
+			}
+		};
+		
+		Collections.sort(listOfIntensities, compareIntensities);
```
tomas-pluskal: What is the purpose of this sorting, and why the new Comparator? (Double is already a Comparable type)

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+		
+		//This is max height for BiGaussian fit. It's in terms of intensities.
+		double maxHeight = listOfIntensities.get(listOfIntensities.size()-1);
+		
+		//Below logic is for finding BiGaussian parameters.		
+		int mu = getScanNumber(maxHeight);
+		int roundedmz = (int) Math.round(mz * 10000); 
+		double halfHeight = (double) maxHeight/2; 
+		biGaussianParams = new double[4];
```
tomas-pluskal: Creating a double[4] object is an awfully C++-like approach. Better to create a class to hold these parameters, and give them proper names. I mean, how are people supposed to know what is stored in biGaussianParams[2]?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+	
+	/**
+	 * <p>
+	 * This method is used for getting scan number for given intensity value.
+	 * </p>
+	 * 
+	 * @param height  a {@link java.lang.Double} object. This is intensity value from the horizontal slice from sparse matrix.
+	 */
+	@SuppressWarnings("rawtypes")
+	private int getScanNumber(double height){
```
tomas-pluskal: What if there are multiple scans with the same intensity?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * <p>
+ * BiGaussian Class is used for fitting bigaussian on EIC.
+ * </p>
+ */
+public class BiGaussian {
+	
+	//This is used to store horizontal slice of sparse matrix.
+	@SuppressWarnings("rawtypes")
+	private final MultiKeyMap horizontalSlice;
```
TDI-tenderholt: Warnings exist for a reason, so I don't think you should be suppressing them without also giving a reason. In this case, I think you should provide types (e.g. ```MultiKeyMap<Double, Double>```?).

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+	 * Inside BiGaussian Constructor we're determining 4 BiGaussian Parameters. MaxHeight, Mu, SigmaLeft and SigmaRight.
+	 * </p>
+	 * 
+	 * @param horizontalSlice a {@link org.apache.commons.collections4.map.MultiKeyMap} object. This is horizontal slice from the sparse matrix.
+	 * @param mz  a {@link java.lang.Double} object. This is m/z value from the raw file.
+	 * @param leftBound a {@link java.lang.Integer} object. This is minimum scan number.
+	 * @param rightBound a {@link java.lang.Integer} object. This is maximum scan number.
+	 */
+	@SuppressWarnings({ "rawtypes", "unchecked" })
+	BiGaussian(MultiKeyMap horizontalSlice,double mz,int leftBound,int rightBound){
```
TDI-tenderholt: Same thing here. Don't suppress warnings and give types to ```MultiKeyMap```.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+import org.apache.commons.collections4.MapIterator;
+import org.apache.commons.collections4.map.MultiKeyMap;
+import org.apache.commons.collections4.keyvalue.MultiKey;
+
+/**
+ * <p>
+ * BiGaussian Class is used for fitting bigaussian on EIC.
+ * </p>
+ */
+public class BiGaussian {
```
TDI-tenderholt: You should put more info in the Javadoc about ```BiGuassian```. The code sounds like it's to handle an asymmetric Gaussian. And now that I think about it, perhaps that's a better class name (i.e. ```AsymmetricGaussian```).

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+	}
+	
+	/**
+	 * <p>
+	 * This method is used for getting scan number for given intensity value.
+	 * </p>
+	 * 
+	 * @param height  a {@link java.lang.Double} object. This is intensity value from the horizontal slice from sparse matrix.
+	 */
+	@SuppressWarnings("rawtypes")
```
TDI-tenderholt: Again, don't suppress warnings without a good justification.

msdk-featdet/msdk-featdet-ADAP-3D/src/test/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussianTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import io.github.msdk.MSDKException;
+import io.github.msdk.datamodel.rawdata.RawDataFile;
+import io.github.msdk.io.netcdf.NetCDFFileImportMethod;
+
+public class BiGaussianTest {
+	
+	  private static final String TEST_DATA_PATH = "src/test/resources/";
```
TDI-tenderholt: You should use class resources instead of a hard-coded path like this. See [```MzMLFileParserTest.java```](https://github.com/msdk/msdk/blob/master/msdk-io/msdk-io-mzml2/src/test/java/io/github/msdk/io/mzml2/MzMLFileParserTest.java) for an example.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+		if(x >= biGaussianParams[1]){
+			exponentialTerm = Math.exp(-1 * Math.pow(x-biGaussianParams[1], 2) / (2 * Math.pow(biGaussianParams[3],2)));
+			biGaussianValue = biGaussianParams[0] * exponentialTerm;
+		}
+		
+		else{
+			exponentialTerm = Math.exp(-1 * Math.pow(x-biGaussianParams[1], 2) / (2 * Math.pow(biGaussianParams[2],2)));
+			biGaussianValue = biGaussianParams[0] * exponentialTerm;
+		}
+
```
TDI-tenderholt: You can make this code simpler, and thus less error prone:

```
double sigma = x >= biGaussianParams[1] ? biGaussianParams[3] : biGaussianParams[2];
double exponentialTerm = Math.exp(-1 * Math.pow(x-biGaussianParams[1], 2) / (2 * Math.pow(sigma,2)));
return biGaussianParams[0] * exponentialTerm;
```

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,198 @@
...
+					interpolationRightSideY2 = triplet2.intensity;
+					break;
+				}
+			}
+				
+		}
+		
+		double interpolationRightSideX = ((halfHeight - interpolationRightSideY1)*(getScanNumber(interpolationRightSideY2)-getScanNumber(interpolationRightSideY1))
+				/(interpolationLeftSideY2 - interpolationRightSideY1))+getScanNumber(interpolationRightSideY1);
+
```
TDI-tenderholt: The two ```for``` loops seem essentially identical except for the iteration bounds (i.e. ```mu-1 .. leftBound``` vs. ```mu+1 .. rightBound```). Suggest refactoring into a single function that takes the bounds you iterate over and returns the ```interpoloated*SideX```.

dharak029:
Hello Dr. Pluskal and Dr. Tenderholt I'm working on your comments. I'll update by EOD.  Thanks for feedback.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: We found two definitions for Asymmetric Gaussian.  http://pubs.acs.org/doi/abs/10.1021/ac60315a005?journalCode=ancham
https://en.wikipedia.org/wiki/Skew_normal_distribution
It's confusing to call it Asymmetric Gaussian. So, we prefer to call it BiGaussian.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: This comparator I've removed.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: I've implemented class for holding these parameters.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: This suggestion helped me to combine two loops in one function and to remove one bug.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: I've added types for this.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: I've implemented a function with one loop only.

msdk-featdet/msdk-featdet-ADAP-3D/src/test/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussianTest.java
dharak029: I've implemented this in all the test cases.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: This made code really simple.

dharak029:
Hello Dr. Pluskal and Dr. Tenderholt,
Your comments were really valuable.  Thanks for that. I've cleaned code. Please take a look

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
TDI-tenderholt: Since BiGaussian is already defined in the literature, I agree that it should stay. Thanks for pointing that out.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,185 @@
...
+	enum Direction {RIGHT, LEFT}
+	
+	//This is used for storing BiGaussian parameters inside constructor.	
+	private static class BiGaussianParams{
+		public static double maxHeight;
+		public static int mu;
+		public static double sigmaLeft;
+		public static double sigmaRight;
+	}
+
```
TDI-tenderholt: I think making these variables static is dangerous. What happens if two difference instances of this class are used by different threads?

I suggest the following, assuming @tomas-pluskal agrees:

```
private static class BiGaussianParams{
  public final double maxHeight;
  public final int mu;
  public final double sigmaLeft;
  public final double sigmaRight;

  public BiGaussianParams(double maxHeight, int mu, double sigmaLeft, double sigmaRight) {
    this.maxHeight = maxHeight;
    this.mu = mu;
    this.sigmaLeft = sigmaLeft;
    this.sigmaRight = sigmaRight;
  }
}
```

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,185 @@
...
+	 * </p>
+	 * 
+	 * @param horizontalSlice a {@link org.apache.commons.collections4.map.MultiKeyMap} object. This is horizontal slice from the sparse matrix.
+	 * @param mz  a {@link java.lang.Double} object. This is m/z value from the raw file.
+	 * @param leftBound a {@link java.lang.Integer} object. This is minimum scan number.
+	 * @param rightBound a {@link java.lang.Integer} object. This is maximum scan number.
+	 */
+	BiGaussian(MultiKeyMap<Integer, Triplet> horizontalSlice,double mz,int leftBound,int rightBound){
+		
+		this.horizontalSlice = horizontalSlice;
```
TDI-tenderholt: Is ```horizontalSlice``` required for the state of this object? Seems like after you've set the ```BiGaussianParams``` at the end of the constructor, you no longer actually need it. If that's true, I suggest removing it—less code/state means easier to maintain and fewer bugs.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,185 @@
...
+	 * </p>
+	 * @param x  a {@link java.lang.Integer} object. This is scan number.
+	 */
+	public double getBiGaussianValue(int x){
+		
+		double sigma = x >= BiGaussianParams.mu ? BiGaussianParams.sigmaRight : BiGaussianParams.sigmaLeft;
+		double exponentialTerm = Math.exp(-1 * Math.pow(x-BiGaussianParams.mu, 2) / (2 * Math.pow(sigma,2)));
+		return BiGaussianParams.maxHeight  * exponentialTerm;
+		
+	}
```
TDI-tenderholt: Note that if you make the variables of ```BiGaussianParams``` non-static, one approach to fix this method would be pass an instance of ```BiGaussianParams``` to this method. If you take that approach, this method can become static since it no longer depends on the state of the ```BiGaussian``` instance. One (major) advantage would be that you can directly unit test this function to ensure its correctness.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,185 @@
...
+		
+		listOfIntensities = new ArrayList<Double>();
+		
+		for(int i=0;i<listOfTripletsInSlice.size();i++){
+			double intensity = listOfTripletsInSlice.get(i)!=null?listOfTripletsInSlice.get(i).intensity:0;
+			listOfIntensities.add(intensity);
+		}
+				
+		Collections.sort(listOfIntensities);
+
```
TDI-tenderholt: I think you can simplify the above code (69-78) using lambdas (each step broken into different lines):
```
List<Double> listOfIntensities = horizontalSlice.values()
  .map(x -> x != null ? x : 0)
  .sorted()
  .collect(Collectors.toList());
```

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,185 @@
...
+				if(triplet2!=null){
+					interpolationY2 = triplet2.intensity;
+					break;
+				}
+			}
+			
+		}
+		
+		double interpolationX =  ((halfHeight - interpolationY1)*((i-previousX2)-i)
+				/(interpolationY2 - interpolationY1))+i;
```
TDI-tenderholt: How likely are you to get a divide-by-zero exception here?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,185 @@
...
+		while(leftBound <= i && i <= rightBound) {
+			
+			i += previousX2;
+			SliceSparseMatrix.Triplet triplet1 = horizontalSlice.get(i,roundedmz);
+			
+			if(triplet1!=null){
+				interpolationY1 = triplet1.intensity;
+			}
+			
+			if(interpolationY1 <halfHeight && triplet1!=null){
```
TDI-tenderholt: If you nest this ```if```-block with the above, you can remove the extra ```triplet1 != null```.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: I need this to find scan number. Hence I can not remove it.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: Can you please explain this in detail?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: Only in one case it can happen. If scan number means i=0 and interpolationY2 and interpolationY1 are same or zero . Should I put it inside try catch?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: This I'm not able to implement. It's giving errors. Anything do I need to include?

TDI-tenderholt: Apologies. I left out the ```.stream()``` and was missing a ```x.intensity```. The following should work:

```
List<Double> listOfIntensities = horizontalSlice.values()
    .stream()
    .map(x -> x != null ? x.intensity : 0.0)
    .sorted()
    .collect(Collectors.toList());
```

Also, ```listOfIntensities``` only appears to be used in the constructor, so I think you can make it a local variable. If you only need the max height, you can do the following:

```
double maxHeight = horizontalSlice.values()
   .stream()
   .map(x -> x != null ? x.intensity : 0.0)
   .max(Double::compareTo)
   .orElse(0.0);
```

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
TDI-tenderholt: Instead of an exception, it's better to test whether ```interpolationY1 == interpolationY1```. Also, isn't the final ```+i``` outside of the division?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
TDI-tenderholt: ```getScanNumber``` is only called in the constructor, right? So you could pass the slice to that method instead of relying on class state.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
TDI-tenderholt: If you change it to the following:

```
public static double getBiGaussianValue(BiGaussianParams params, int x){
		
	double sigma = x >= params.mu ? params.sigmaRight : params.sigmaLeft;
	double exponentialTerm = Math.exp(-1 * Math.pow(x-params.mu, 2) / (2 * Math.pow(sigma,2)));
	return params.maxHeight  * exponentialTerm;
		
	}
```

Then in a unit test, you can do something like:

```
BiGaussianParams params;
params.mu = 30000;
params.maxHeight = 2e6;
params.sigmaLeft = 10;
params.sigmaRight = 20;

assertThat(BiGaussian.getValue(params, 30000), equalTo(2e6));
assertThat(BiGaussian.getValue(params, 29900), equalTo(....));
assertThat(BiGaussian.getValue(params, 30010), equalTo(....));
```

(Note that I've shorted the method from ```getBiGaussianValue``` to ```getValue``` since it's pretty clear the value is for a BiGaussian, and I've used ```equalTo``` from Hamcrest Matchers because I can never remember the parameter order for ```assertEquals```.)

In contrast, to test the current ```getBiGaussianValue()```, you actually have to feed in real data to the ```BiGaussian``` constructor.

Make sense?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: Yes.Final +1 is outside the division.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: Yes. That can also work.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
tomas-pluskal: I would say it might be simpler to just make the maxHeight, mu, sigmaLeft and sigmaRight class variables of BiGaussian class and remove the BiGaussianParams class completely (unless Dharak plans to add more functionality to it).

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: This is done. Thanks:)

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: This is done.Thanks:)

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: No I'm not planning to add more functionality. Hence I've made them class variables.

TDI-tenderholt: What other functionality will you add? Do these four values need to be mutable? If not, you should make them ```final```.

dharak029: For single object of BiGaussian these values are immutable.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: I've removed BigaussianParams class as @tomas-pluskal  suggested.

dharak029:
@TDI-tenderholt and @tomas-pluskal thanks for all the inputs. I've implemented the changes you suggested. The code looks more compact and readable. Please have a look.

tomas-pluskal:
Hi Dharak,
One more comment: if you're using Eclipse, could you install the Google Code style (eclipse-java-google-style.xml from https://github.com/google/styleguide) and reformat your source codes using this style (in Eclipse menu Source / Format). Currently your code looks very different from the rest of MSDK, using tabs for indentation etc.

TDI-tenderholt:
Yes, I think it's much better, although you should handle the possible divide by zero in ```InterpolationX```. What's a reasonable value in this case? 

Also, it's still a bit hard understand what is happening in ```InterpolationX```, but that might just be me. Are all of the variables named appropriately? For instance, ```halfHeight``` sounds like an intensity value while its Javadoc description sounds like it's a m/z value. Similarly, ```roundedmz``` sounds like a m/z value, but it's calculated from a scan number (i.e. ```int roundedmz = (int) Math.round(mz * 10000)```).

Finally, you should remove the unused imports.

--
P.S. Unrelated to the PR, I ran FindBugs and it looks like there's a bug in ```Peak3DTest```:

```
if(curMZ == null && Math.abs(curMZ - roundedMz) >= 2*roundedFWHM)
   break;
```

If the first part is true, then won't the second part throw a null pointer exception when you try to subtract something from ```curMZ```?

dharak029:
@tomas-pluskal  I formatted code with Google code style. Please have a look.

dharak029:
@TDI-tenderholt apologies for wrong java doc info. Yes half height is half of  maximum intensity value which is maxHeight. I'll also resolve rest of the bugs which you found.

tomas-pluskal:
Hi Dharak,
Please make sure you add the explanation to the comments in the code (it is more important to have it there, than here).

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
```diff
@@ -28,117 +28,125 @@
...
+        matrixIndex++;
+      }
+
+      try {
+        double[] parameters = GaussianCurveFitter.create().fit(obs.toList());
+        // createDataFile(verticalSlice,parameters,randInt);
+        sigma += 2.35482 * parameters[2];
+
+      } catch (Exception e) {
+        continue;
```
tomas-pluskal: Do not swallow exceptions

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
```diff
@@ -28,117 +28,125 @@
...
+        buffer.append(datapoint.get(i).mz).append(",").append(datapoint.get(i).intensity)
+            .append(",").append(gaussianY[i]).append("\r\n");
+        writer.write(buffer.toString());
+        buffer = new StringBuffer();
+      }
+      writer.close();
+    }
+
+    catch (Exception e) {
+
```
tomas-pluskal: Do not swallow exceptions.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
```diff
@@ -28,117 +28,125 @@
...
-			writer.close();
-		}
-		
-		catch(Exception e){
-			
-		}
-		
-	}
+
+  private static final String TEST_DATA_PATH = "src/test/resources/";
```
tomas-pluskal: I think it was mentioned before - let's use getClass.getClassLoader.getResource() instead of hard-coding the path. Hard-coding the path fails in some environments.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
dharak029: Can you please explain this?

TDI-tenderholt: I also think you should catch specific exceptions and handling each appropriately.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
```diff
@@ -28,117 +28,125 @@
...
+   */
+  public double estimateFwhmMs(int numberOfScansForFWHMCalc) {
+
+    double sigma = 0;
+    int countProperIteration = 0;
+    int countTotalIteration = 0;
+
+    while (countProperIteration < numberOfScansForFWHMCalc) {
+      countTotalIteration++;
+      int matrixIndex = 0;
```
TDI-tenderholt: Eclipse says ```matrixIndex``` isn't actually used. If that's the case, remove it.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
dharak029: Ohh.Ok. Thanks.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
TDI-tenderholt: Few other comments:

* Use try-with resources:

```
try(BufferedWriter writer = new BufferedWriter(new FileWriter(TEST_DATA_PATH + filename))) {
  ...
} catch (Exception e) {
  ..
}
```

* Don't hard code a platform-specific new line (i.e. ```\r\n```) when you can use ```BufferedWriter.newLine()```.

* If you don't actually need the ```gaussianY``` array, don't create it. You should be able to just use a single ```double``` at each place in the loop.

dharak029: I've removed that function as I'm not using that.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
```diff
@@ -0,0 +1,184 @@
...
+   *        slice from sparse matrix.
+   */
+  private int getScanNumber(MultiKeyMap<Integer, Triplet> horizontalSlice, double height) {
+    int mu = 0;
+    MapIterator<MultiKey<? extends Integer>, Triplet> iterator = horizontalSlice.mapIterator();
+
+    while (iterator.hasNext()) {
+      iterator.next();
+
+      MultiKey<? extends Integer> mk = (MultiKey<? extends Integer>) iterator.getKey();
```
TDI-tenderholt: I don't think you need the explicit ```(MultiKey<? extends Integer>)``` cast.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
```diff
@@ -15,130 +15,78 @@
...
+      for (SliceSparseMatrix.VerticalSliceDataPoint datapoint : verticalSlice) {
+        obs.add(datapoint.mz, datapoint.intensity);
+      }
+
+      try {
+        double[] parameters = GaussianCurveFitter.create().fit(obs.toList());
+        sigma += 2.35482 * parameters[2];
+
+      } catch (MathIllegalArgumentException e) {
+        continue;
```
TDI-tenderholt: As @tomas-pluskal said before, don't just swallow this exception. If this exception is thrown, is it fatal and you should throw a custom ```MSDKException``` or can you add a default value to ```sigma```?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/SortAndKeepOriginalIndecies.java
```diff
@@ -15,50 +15,52 @@
...
+
+  /** {@inheritDoc} */
+  @Override
+  public int compare(Integer index1, Integer index2) {
+    if (dataArr[index2] > dataArr[index1]) {
+      return -1;
+    } else if (dataArr[index2] < dataArr[index1]) {
+      return 1;
+    } else {
+      return 0;
```
TDI-tenderholt: I think you can simplify:

```
@Override
public int compare(Integer index1, Integer index2) {
  return Double.compare(dataArr[index2], dataArr[index1])
}
```

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
dharak029: it was giving me warnings so for removing it I did that. Should I remove it?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
dharak029: The loop should be continued if it throws the exception that's why I have not put any value of sigma. It is not fatal.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
TDI-tenderholt: That's odd. Eclipse 4.6.3 doesn't show an warning. What IDE are you using and warning are you getting?

dharak029: Unchecked cast from MultiKey<capture#2-of ? extends Integer> to MultiKey<Integer>.
I'm using eclipse 4.6.3 neon

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
TDI-tenderholt: Ok, fine by me then.

Also, as I'm looking at it again, perhaps the ```2.35482``` should be set as a class constant. It's the 99% confidence interval, right?

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/BiGaussian.java
TDI-tenderholt: Weird. I guess you can keep the cast then.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/SortAndKeepOriginalIndecies.java
dharak029: I tried to change it but it's giving me errors in test. This code was written by Owen for continuous wavelet transform.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/CurveTool.java
dharak029: Yeah. Okay let me make it constant.

msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D/common/algorithms/SortAndKeepOriginalIndecies.java
TDI-tenderholt: Switch the order of the indices? ```Double.compare(dataArr[index1], dataArr[index2]);```.

dharak029: Yeah. It's working.Thanks:)

dharak029:
I hope now the code looks fine and it gets merge. haha :)

TDI-tenderholt:
Yes, the PR did drag on as we iterated through the review. Part of it was because more and more unrelated changes were made (i.e. source format of other files). Ideally send smaller changes for review.

I'll let @tomas-pluskal do the actual merging of the PR when he's ready.

PR #227, opened 2017-07-23
Predicate for Scans & Chromatograms, Import from InputStreams

adhi1419:


tomas-pluskal:
The methods like 
````
public double[] getMzValues(double array[])
````
take an array[] as a parameter. The purpose of this is that you should load the data into a given array, as long as the array is large enough to hold all the values. If the array is not large enough, create a new array (because arrays cannot be resized in Java). The returned array is either the same as the provided one or the new one. The purpose of this logic is to avoid creating tons of array objects that need to be collected by the garbage collector.

tomas-pluskal:
Could you also add a jUnit test that will call getMzValues() or getIntensityValues on a scan that was not parsed, and therefore generate an exception?

adhi1419:
Currently, the list of `MsScan`s and `Chromatogram`s contained in the RawFile only has the scans that pass the Predicate. 
I'll add all scans to this list instead.

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #232, opened 2017-08-07
Add random access feature to existing netCDF parser

adhi1419:


msdk-io/msdk-io-netcdf/src/main/java/io/github/msdk/io/netcdf/NetCDFFileImportMethod.java
```diff
@@ -82,7 +85,12 @@
    * @param dataStore a {@link io.github.msdk.datamodel.datastore.DataPointStore} object.
    */
   public NetCDFFileImportMethod(@Nonnull File sourceFile) {
+    this(sourceFile, null);
+  }
+
+  public NetCDFFileImportMethod(@Nonnull File sourceFile, Predicate<MsScan> msScanPredicate) {
     this.sourceFile = sourceFile;
+    this.msScanPredicate = msScanPredicate != null ? msScanPredicate : s -> true;
```
TDI-tenderholt: Can you avoid explicitly passing in a ```null``` value? What about initializing ```msScanPredicate``` to ```s -> true``` (line 68) and then using the ```Predicate.and()``` method?

msdk-io/msdk-io-netcdf/src/main/java/io/github/msdk/io/netcdf/NetCDFMsScan.java
```diff
@@ -0,0 +1,154 @@
...
+    this.intensityValueVariable = intensityValueVariable;
+    this.massValueScaleFactor = massValueScaleFactor;
+    this.intensityValueScaleFactor = intensityValueScaleFactor;
+    this.mzValues = null;
+    this.intensityValues = null;
+    this.spectrumType = null;
+  }
+
+  @Override
+  public float[] getIntensityValues(float[] intensityValues) {
```
tomas-pluskal: If the Predicate was set for the scan, this function should return the pre-loaded data

msdk-io/msdk-io-netcdf/src/main/java/io/github/msdk/io/netcdf/NetCDFMsScan.java
```diff
@@ -54,51 +54,70 @@ public NetCDFMsScan(Integer scanNumber, int[] scanStartPositions, float[] scanRe
...
+        for (int i = 0; i < numOfDataPoints; i++) {
+          final Index intensityIndex0 = intensityValuesIndex.set0(i);
+          intensityValues[i] =
+              (float) (intensityValueArray.getDouble(intensityIndex0) * intensityValueScaleFactor);
+        }
+      } catch (IOException | InvalidRangeException e) {
+        throw new MSDKRuntimeException(e);
+      }
+    } else {
       if (intensityValues == null || intensityValues.length < numOfDataPoints)
```
tomas-pluskal: is the intensityValues == null condition here necessary?

msdk-io/msdk-io-netcdf/src/main/java/io/github/msdk/io/netcdf/NetCDFMsScan.java
```diff
@@ -54,51 +54,70 @@ public NetCDFMsScan(Integer scanNumber, int[] scanStartPositions, float[] scanRe
...
   public float[] getIntensityValues(float[] intensityValues) {
-    final Integer scanIndex = getScanIndex();
-    numOfDataPoints = getNumberOfDataPoints();
-    try {
-      final int scanStartPosition[] = {scanStartPositions[scanIndex]};
-      final int scanLength[] = {scanStartPositions[scanIndex + 1] - scanStartPositions[scanIndex]};
-      final Array intensityValueArray = intensityValueVariable.read(scanStartPosition, scanLength);
-      final Index intensityValuesIndex = intensityValueArray.getIndex();
-
+    if (this.intensityValues == null) {
```
tomas-pluskal: maybe better to rename this.intensityValues to something like this.preloadedIntensityValues, for clarity

msdk-io/msdk-io-netcdf/src/main/java/io/github/msdk/io/netcdf/NetCDFMsScan.java
adhi1419: The `getIntensityValues()` function calls this function as `getIntensityValues(null)`. So, it is necessary to check if the passed array is null before checking if the data fits into it.

PR #235, opened 2017-08-09
Use TIC value calculated using the MsSpectrumUtil

adhi1419:
…available in the mzml file


msdk-io/msdk-io-mzml2/src/main/java/io/github/msdk/io/mzml/MzMLFileExportMethod.java
```diff
@@ -238,15 +238,20 @@ public Void execute() throws MSDKException {
...
-            writeCVParam(xmlStreamWriter,
-                new MzMLCVParam(MzMLCV.cvTIC, String.valueOf(tic), "total ion current", null));
-          }
+        if (scan.getTIC() != null) {
+          Float tic = scan.getTIC();
+          writeCVParam(xmlStreamWriter,
+              new MzMLCVParam(MzMLCV.cvTIC, String.valueOf(tic), "total ion current", null));
         }
 
+        // remove tic value read from the file
```
tomas-pluskal: If you want to remove the TIC CV term from the existing MzMLRawDataFile file, it makes more sense to remove it during importing the file, not during exporting a new file (it makes the source file inconsistent, because its CVParam list would change after exporting)

tomas-pluskal:
There seems to be a test failure in the msdk-io-mzml module

adhi1419:
The test now fails at `SrmDetectionMethodTest` now (I fixed the other errors), I do not think it has anything to with the mzml module

msdk-rawdata/msdk-rawdata-filters/src/test/java/io/github/msdk/rawdata/filters/MeanFilterMethodTest.java
```diff
@@ -117,9 +117,10 @@ public void testMeanFilter() throws Exception {
       float intensityValues[] = newScan.getIntensityValues();
 
       float intensityAverage = inputScan.getTIC() / inputScan.getNumberOfDataPoints();
+      System.out.println(inputScan.getTIC());
 
       for (float intValue : intensityValues) {
-        Assert.assertEquals(intensityAverage, intValue, 0.001);
+        Assert.assertEquals(intensityAverage, intValue, 0.4);
```
tomas-pluskal: This looks suspicious - with the previous mzML parser, the test passed with 0.001 threshold. Why is it now changed to 0.4?

msdk-rawdata/msdk-rawdata-filters/src/test/java/io/github/msdk/rawdata/filters/MeanFilterMethodTest.java
```diff
@@ -117,9 +117,10 @@ public void testMeanFilter() throws Exception {
       float intensityValues[] = newScan.getIntensityValues();
 
       float intensityAverage = inputScan.getTIC() / inputScan.getNumberOfDataPoints();
+      System.out.println(inputScan.getTIC());
```
tomas-pluskal: Remove

msdk-io/msdk-io-mzml/src/test/java/io/github/msdk/io/mzml/MzMLFileExportMethodTest.java
```diff
@@ -246,7 +246,7 @@ public void testFileWithUV() throws MSDKException, IOException {
         scan.getNumberOfDataPoints());
     Assert.assertEquals(Integer.valueOf(2114), scan.getScanNumber());
     Assert.assertEquals(MsSpectrumType.CENTROIDED, scan.getSpectrumType());
-    Assert.assertEquals(9.939699e06f, scan.getTIC(), 0.0001);
+    Assert.assertEquals(9.939699e06f, scan.getTIC(), 1e01);
```
tomas-pluskal: I don't think you should be changing the thresholds here

msdk-io/msdk-io-mzml/src/test/java/io/github/msdk/io/mzml/MzMLFileImportMethodTest.java
```diff
@@ -78,7 +78,7 @@ public void testFileWithUV() throws MSDKException {
         scan.getNumberOfDataPoints());
     Assert.assertEquals(Integer.valueOf(2114), scan.getScanNumber());
     Assert.assertEquals(MsSpectrumType.CENTROIDED, scan.getSpectrumType());
-    Assert.assertEquals(9.939699e06f, scan.getTIC(), 0.0001);
+    Assert.assertEquals(9.939699e06f, scan.getTIC(), 1e01);
```
tomas-pluskal: I don't think you should be changing the thresholds here

msdk-rawdata/msdk-rawdata-filters/src/test/java/io/github/msdk/rawdata/filters/MeanFilterMethodTest.java
adhi1419: The range of the error increased for reasons I am not sure of. It might have something to do with the intensity values decoded, but, I am not certain why.

msdk-io/msdk-io-mzml/src/test/java/io/github/msdk/io/mzml/MzMLFileExportMethodTest.java
adhi1419: The expected value is what is found in the mzML file. The computed value surprisingly has an error of several magnitudes higher.
This seems to be a recurring issue with the parsed intensity values, and might have to do something with the conversion from 64-bit to 32-bit precision. I will have a look into the float decoder to make sure there isn't a problem with it.

tomas-pluskal: OK, I guess it is not so surprising to get this kind of error when using floats. But the issue of the mean filter module should be something else.

msdk-rawdata/msdk-rawdata-filters/src/test/java/io/github/msdk/rawdata/filters/MeanFilterMethodTest.java
```diff
@@ -119,7 +119,7 @@ public void testMeanFilter() throws Exception {
       float intensityAverage = inputScan.getTIC() / inputScan.getNumberOfDataPoints();
 
       for (float intValue : intensityValues) {
-        Assert.assertEquals(intensityAverage, intValue, 0.001);
+        Assert.assertEquals(intensityAverage, intValue, 0.4);
```
tomas-pluskal: The old mzML reader calculated the TIC values using the same MsSpectrumUtil.getTIC() function. And this test passed using the old mzML reader. Please try to find out why this is happening.

adhi1419: I changed the logic in the decoder to read float/double from the stream directly, but, the condition doesn't seem much (or change at all, which is actually as expected).
I am unable to locate the origin of this loss in precision in the new parser.
I ran the same filtering test on other mzML files and the delta reduced up to 0.01.
I tried using jmzml's logic of using `floatValue()` on the `Double` object read from the stream instead of casting it to double, but, it made no difference.

tomas-pluskal: So if you read with the old parser and with the new parser, and compare the arrays returned by getIntensityValues[], the values in the arrays are different?
I am sure you can find the cause if you dig a bit deeper into the issue, add some more debugging output, etc.

adhi1419: I compared the intensity arrays of both the new and the old parser, and the values matched one another. They were exactly the same.

tomas-pluskal: Ok, so what is the reason that the test now fails? Are the getTIC() values different?

adhi1419: Yes, the getTIC() values are (slightly) different only for this file.
I checked it for other files, and the TIC value calculated didn't vary as much.

tomas-pluskal: Even a small deviation of intensity values will add up when calculating the TIC. We need to find where the deviation comes from.

adhi1419: I noticed one more thing, when I checked the intensity values of the original unfiltered scans, they were filtered too, meaning, the original intensity values were not retained (Which is not the case with the old parser, obviously). So, I think it is safe to assume that the error is actually arising from handling the instances of the intensity arrays wrongly. I will try locating the origin of the error and get back here.

tomas-pluskal:
Sounds like a refererence integrity issue, as I was pointing out before.
Try to find it.

tomas-pluskal:
What is the situation regarding the returned getTIC() values?

adhi1419:
@tomas-pluskal In the JUnit tests for the mzML import method, there still is a deviation from the TIC value found in the file. I think we can attribute it to the conversion from double to float values.
In the filtering method, fixing the instance bug, resolved the issue with the increase in the delta for that particular assertion which was failing.

tomas-pluskal:
OK, so can you change the assert back to 
````
Assert.assertEquals(intensityAverage, intValue, 0.001);
````
?

File: msdk-rawdata/msdk-rawdata-filters/src/test/java/io/github/msdk/rawdata/filters/MeanFilterMethodTest.java [hunk 7]
```diff
@@ -119,7 +119,7 @@ public void testMeanFilter() throws Exception {
       float intensityAverage = inputScan.getTIC() / inputScan.getNumberOfDataPoints();
 
       for (float intValue : intensityValues) {
-        Assert.assertEquals(intensityAverage, intValue, 0.4);
+        Assert.assertEquals(intensityAverage, intValue, 0.0001);
```
adhi1419: I forced pushed this change.

PR #245, opened 2017-09-15
Added parameters to ADAP3DFeatureDetectionMethod

asmirn1:
Renamed class Parameters and added it as an argument of ADAP3DFeatureDetectionMethod


msdk-featuredetection-adap3d/README.md
```diff
@@ -6,15 +6,15 @@ ADAP-3D was originally developed by the Du-Lab research team ([http://www.du-lab
...
 
 ## Useful Link
 
 1. [Link to Commits](https://github.com/msdk/msdk/commits?author=dharak029)
 2. [Link to Pull Requests](https://github.com/msdk/msdk/pulls?q=is%3Apr+is%3Aclosed+no%3Aassignee+author%3Adharak029)
-3. [Link to Code](https://github.com/msdk/msdk/tree/master/msdk-featuredetection-adap3d/src/main/java/io/github/msdk/featdet/ADAP3D)
-4. [Link to TestCases](https://github.com/msdk/msdk/tree/master/msdk-featuredetection-adap3d/src/test)
-5. [Link to Detailed Project Report](https://github.com/msdk/msdk/blob/master/msdk-featuredetection-adap3d/ADAP3D%20Project%20Report.docx)
+3. [Link to Code](https://github.com/msdk/msdk/tree/master/msdk-featdet/msdk-featdet-ADAP-3D/src/main/java/io/github/msdk/featdet/ADAP3D)
```
tomas-pluskal: I guess this commit accidentally reverts the correct links back to the old ones

msdk-featuredetection-adap3d/src/main/java/io/github/msdk/featdet/ADAP3D/ADAP3DFeatureDetectionMethod.java
```diff
@@ -160,15 +165,15 @@ public ADAP3DFeatureDetectionMethod(@Nonnull RawDataFile rawFile, @Nullable Pred
         Collections.max(peakWidthList) + LOW_BOUND_PEAK_WIDTH_PERCENT * avgPeakWidth;
 
     // set the estimated parameters.
-    objParameters.setLargeScaleIn(highestWaveletScale);
-    objParameters.setMinPeakWidth(peakDurationLowerBound);
-    objParameters.setMaxPeakWidth(peakDurationUpperBound);
-    objParameters.setCoefAreaRatioTolerance(coefOverAreaThreshold);
+    parameters.setLargeScaleIn(highestWaveletScale);
```
tomas-pluskal: I see a problem here that the parameter values are always overwritten with the estimated ones - even if the user provided pre-defined parameters in the constructor. It should probably be implemented in a different way - the default value for all the parameters should be null, and user can set some of them. Those that are not provided by the user should be set during the estimation stage.

msdk-featuredetection-adap3d/README.md
asmirn1: Ah, okay. I though I had the latest changes but maybe I didn't. I'll fix it.

msdk-featuredetection-adap3d/src/main/java/io/github/msdk/featdet/ADAP3D/ADAP3DFeatureDetectionMethod.java
asmirn1: Our idea was that user would set up initial parameters (or use the default values), and than those parameters are estimated from the first detected peaks. But I see your point. I need to discuss it to Xiuxia.

PR #248, opened 2017-11-23
Do not skip MzMLCV.cvTIC

nilshoffmann:
Some vendors report a different TIC CV param value from the one obtained from integrating from the scan's intensities.
This may be problematic if the TIC CV param value is the result e.g. data processing and should be used for downstream analysis.
Retain the MzMLCV.cvTIC so it can be retrieved from the MzMLMsScan object manually, if necessary.


tomas-pluskal:
Thanks, Nils. I agree we can keep it. The question is what to do when the file is exported - should we add the calculated TIC or the saved one? Especially consider that the raw intensities might have changed by processing.

Issue #246, opened 2017-09-19
ADAP3D issues


tomas-pluskal:
This issue is for tracking the development and bugs of ADAP3D.

1. Any of these files produce exceptions or hang the algorithm: [test_files.zip](https://github.com/msdk/msdk/files/1313783/test_files.zip)

2. The tiny.mzXML file that is included in the ADAP3D test suite contains a very obvious peak at 166.071 m/z, 2.9 min. that is not found by the algorithm:

![shot](https://user-images.githubusercontent.com/1796692/30587218-5fadcac4-9d00-11e7-86dd-c3058982a1f7.png)

3. The cancel() method does not work if the algorithm is stuck at the FWHM estimation step

tomas-pluskal:
4. Running ADAP3D on the DC_010814_StandardsMixTest1_34StandardMix_01.mzXML file produces some duplicated peaks, like here: 

![607mz](https://user-images.githubusercontent.com/1796692/30830717-f8cbcbf4-a212-11e7-9947-ee33b39e42c2.png)

Aniket-Pradhan:
Hey there.
Can you please explain the steps you took to execute the algorithm using the test files. 
I have been looking at the ADAP 3D classes, but unable to find any driver class.

asmirn1:
Hi Aniket,

The tests above have been executed through [MZmine2](https://github.com/mzmine/mzmine2/). However, ADAP3D is still being developed and not included into any of the official MZmine2 releases. If you are interested in trying ADAP3D, we might be able to provide a developer version of MZmine2 with ADAP3D enabled.

Aniket-Pradhan:
Hi Asmirn1,

It would be great if you could provide me the developer version of MZmine2 with ADAP3D, so that I can look more into the issue, and help in solving it.

du-lab:
Hi Asmirn1,

We will be glad to provide the developer version of ADAP-3D to you. But please keep in mind that this version is just the first draft of the algorithm and might be changed considerably in the near future.

May I ask how you plan to solve the issue?

asmirn1:
MZmine2 with enabled ADAP3D algorithm can be found [here](https://github.com/du-lab/mzmine2/tree/with_adap3d) on the branch *with_adap3d*. To run the algorithm, go to the menu `Raw data methods -> Peak detection -> ADAP3D feature detection (experimental)`. Let me know if you have any questions.

Aniket-Pradhan:
Hi @du-lab ,

[...]

PR #253, opened 2018-03-08
Optimized code for BiGaussian class #246

asmirn1:
Mostly, we removed the unnecessary binary search used in a loop. Also, we simplified the code and removed some extra variables.


tomas-pluskal:
Awesome, looking forward to testing it!

PR #255, opened 2018-05-14
Add sirius support

evgerher:
The code right now does not do a lot. It does only getting familiar with Sirius API.
The rest will be done in future commits.


tomas-pluskal:
Move the query.txt to src/test/resources

msdk-id-sirius/src/main/java/ISirius.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Scanner;
+import java.util.stream.Stream;
+
+/**
+ * Created by evger on 14-May-18.
+ */
+
+public class ISirius {
+
+    public static Pair<double[], double[]> readMsFile(String path) throws FileNotFoundException {
```
tomas-pluskal: Instead of this, use MspImportAlgorithm from the msdk-io-msp package. MSP is a simple text format for mass spectra.

msdk-id-sirius/src/main/java/ISirius.java
```diff
@@ -0,0 +1,73 @@
...
+        Double mzArray[] = mz.toArray(new Double[mz.size()]);
+        Double intensiveArray[] = intensive.toArray(new Double[intensive.size()]);
+
+        double[] mzPrimitive = Stream.of(mzArray).mapToDouble(Double::doubleValue).toArray();
+        double[] intensivePrimitive = Stream.of(intensiveArray).mapToDouble(Double::doubleValue).toArray();
+
+        return new Pair<>(mzPrimitive, intensivePrimitive);
+    }
+
+    public static void main(String[] args) {
```
tomas-pluskal: There should be no main() function here. Instead, you can start writing simple junit tests in the src/test/java folder.
If you have no experience with junit, read some tutorials first.

msdk-id-sirius/pom.xml
```diff
@@ -9,40 +9,105 @@
...
-      <id>bioinf-jena-sirirusCli</id>
+      <id>bioinf-jena-main</id>
+      <name>bioinf-jena-release</name>
+      <url>https://bio.informatik.uni-jena.de/repository/libs-release-local</url>
+    </repository>
+    <repository>
+      <id>bioinf-jena</id>
       <name>bioinf-jena-release</name>
-      <url>http://bio.informatik.uni-jena.de/artifactory/libs-releases-local</url>
+      <url>https://bio.informatik.uni-jena.de/repository/local</url>
```
tomas-pluskal: Do you need this local repository? Shouldn't all the relevant artifacts be in the libs-release-local repo?

chhh: The three repos required to build were taken from: https://github.com/boecker-lab/sirius/blob/release/build.gradle

chhh: Relevant portion:
```groovy
repositories {
        mavenLocal()
        mavenCentral()
        maven {
            url "https://bio.informatik.uni-jena.de/repository/local"
        }
        maven {
            url "http://www.ebi.ac.uk/Tools/maven/repos/content/groups/ebi-repo/"
            //url "http://www.ebi.ac.uk/intact/maven/nexus/content/repositories/ebi-repo"
        }
        maven {
            url "https://www.xypron.de/repository" //glpk repo
        }
    }
```

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,95 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Created by evger on 14-May-18.
+ */
+
+public class SiriusIdentificationMethod implements MSDKMethod<IonAnnotation> {
+  Sirius sirius;
+
+  SiriusIdentificationMethod() {
```
tomas-pluskal: make constructor public, and add all relevant input parameters of the method to the constructor

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,95 @@
...
+      if (splitted.length == 2) {
+        mz[index] = Double.parseDouble(splitted[0]);
+        intensive[index++] = Double.parseDouble(splitted[1]);
+      } else throw new MSDKRuntimeException("Incorrect spectrum structure");
+    }
+
+    return new Pair<>(mz, intensive);
+  }
+
+  public List<IdentificationResult> identifyMs2Spectrum(
```
tomas-pluskal: move the logic to the execute() method, remove identifyMs2Spectrum

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,95 @@
...
+      @Nullable Pair<double[], double[]> ms1pair, Pair<double[], double[]> ms2pair,
```
tomas-pluskal: Remove the Pair class, the method should just accept MsSpectrum objects as parameters

msdk-id-sirius/src/test/java/SiriusMs2Test.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.Test;
+
+/**
+ * Created by evger on 15-May-18.
+ */
+public class SiriusMs2Test {
+
+  @Test
+  public void testCreateMs2ExperimentMsp() throws MSDKException, IOException {
+    String ms2MspPath = "target/test-classes/sample.msp";
```
tomas-pluskal: Change this to use the class resources, like I commented in your MGF module

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.List;
+import java.util.Scanner;
+import javax.annotation.Nullable;
+
+/**
+ * Created by evger on 14-May-18.
+ */
+
+public class SiriusIdentificationMethod implements MSDKMethod<IonAnnotation> {
+  Sirius sirius;
```
tomas-pluskal: if this variable never changes, make it final. always use final when possible: http://www.javapractices.com/topic/TopicAction.do?Id=23

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,188 @@
...
+    this.parentMass = parentMass;
+    this.ion = ion;
+    numberOfCandidates = 5;
+  }
+
+  public double getParentMass() {
+    return parentMass;
+  }
+
+  public void setParentMass(double mass) {
```
tomas-pluskal: I don't think you should have setters for these fields - what if someone changes the value in the middle of the processing? The results would be inconsistent. Instead, make the fields final.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,188 @@
...
+  private final Sirius sirius;
+  private MsSpectrum ms1;
+  private MsSpectrum ms2;
+  private double parentMass;
+  private String ion;
+  private int numberOfCandidates;
+  private List<IonAnnotation> result;
+
+  public SiriusIdentificationMethod(@Nullable MsSpectrum ms1, MsSpectrum ms2, double parentMass,
+      String ion) {
```
tomas-pluskal: Use the MSDK's IonType instead of String for ion.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,188 @@
...
+
+  private final Sirius sirius;
+  private MsSpectrum ms1;
+  private MsSpectrum ms2;
+  private double parentMass;
+  private String ion;
+  private int numberOfCandidates;
+  private List<IonAnnotation> result;
+
+  public SiriusIdentificationMethod(@Nullable MsSpectrum ms1, MsSpectrum ms2, double parentMass,
```
tomas-pluskal: Please use Nonnull for ms2, and change parentMass from double to @Nonnull Double. 
It is better to avoid using primitive types unless it is for performance reasons (like large data arrays).

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,188 @@
...
+  @Override
+  public Float getFinishedPercentage() {
+    return null;
+  }
+
+  @Nullable
+  @Override
+  public List<IonAnnotation> execute() throws MSDKException {
+    result = new ArrayList<>();
+    List<IdentificationResult> siriusSpectrums = siriusProcessSpectrums();
```
tomas-pluskal: By the way, the plural of "spectrum" is "spectra"

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,188 @@
...
+    return result;
+  }
+
+  @Nullable
+  @Override
+  public List<IonAnnotation> getResult() {
+    return result;
+  }
+
+  // TODO: implement
```
tomas-pluskal: Go ahead and implement it

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+
+  /**
+   * Transformation of MSDK data structures into Sirius structures and processing by sirius
+   */
+  protected List<IdentificationResult> siriusProcessSpectrums() throws MSDKException {
+    Spectrum<Peak> siriusMs1 = null, siriusMs2;
+    siriusMs2 = sirius
+        .wrapSpectrum(ms2.getMzValues(), LocalArrayUtil.convertToDoubles(ms2.getIntensityValues()));
```
tomas-pluskal: It's a bad practice to have many function calls on a single line, because when an exception occurs on that line, you won't know where it came from. It's better to split this into several lines (first get ms2.getMzValues(), then get the intensity values, then call sirius.wrapSpectrum...)

TDI-tenderholt: An alternative solution is for `sirius.wrapSpectrum` to take just `ms2`, and then call `ms2.getMzValues()` and `LocalArrayUtil.convertToDoubles(ms2.getIntensityValues())`.

tomas-pluskal: We cannot change sirius, is's coming from an imported library.

msdk-id-sirius/src/main/java/NativeLibraryLoader.java
```diff
@@ -0,0 +1,140 @@
...
+    try {
+      return Paths.get(url.toURI()).toAbsolutePath();
+    } catch (URISyntaxException e) {
+      throw new MSDKException(e);
+    }
+  }
+
+  public static void loadLibraryFromJar(String folder, String[] libs) throws MSDKException, IOException {
+    String suffix = "", realPath;
+    String arch = System.getProperty("os.arch").toLowerCase().endsWith("64") ? "w64" : "w32";
```
tomas-pluskal: Check here for the possible os.name and os.arch values: http://www.java-gaming.org/index.php/topic,14110

msdk-id-sirius/src/main/java/NativeLibraryLoader.java
```diff
@@ -0,0 +1,140 @@
...
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+
+public class NativeLibraryLoader {
+  private NativeLibraryLoader() {}
```
tomas-pluskal: Add empty lines between methods. Use the Google code style for formatting.

msdk-id-sirius/pom.xml
```diff
@@ -43,36 +43,36 @@
     <dependency>
       <groupId>de.unijena.bioinf.ms</groupId>
       <artifactId>chemistry_base</artifactId>
-      <version>4</version>
+      <version>4.0</version>
```
chhh: Please move this version number to a property in pom.xml. I've posted a full example repo for you which both demonstrates the use of a property and how to use slf4j with various logging backends: https://github.com/chhh/slf4j-example

msdk-id-sirius/pom.xml
```diff
@@ -123,7 +123,11 @@
       <artifactId>cdk-silent</artifactId>
       <version>${cdk.version}</version>
     </dependency>
-
+    <dependency>
```
chhh: This dependency is already specified in the parent pom, you don't need it.

evgerher: cdk-silent dependency exists only in msdk-io-mztab & msdk-isotopes-isotopepattern, not in parent pom
The code uses a SilentChemObjectBuilder, which can be found only in cdk-silent package

msdk-id-sirius/pom.xml
```diff
@@ -6,4 +6,131 @@
     <version>0.0.12-SNAPSHOT</version>
   </parent>
   <artifactId>msdk-id-sirius</artifactId>
+
+  <properties>
+    <sirius.version>4.0</sirius.version>
+    <slf4j.version>1.7.25</slf4j.version>
```
tomas-pluskal: This property is already defined in the parent pom.xml

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,349 @@
...
+ * Transformation of IdentificationResult (Sirius) into IonAnnatation (MSDK)
+ * Containts sub-class for generating FormulaConstraints for Sirius from MolecularFormulaRange object
+ */
+public class SiriusIdentificationMethod implements MSDKMethod<List<IonAnnotation>> {
+
+  /**
+   * Dynamic loading of native libraries
+   */
+  static {
+    try {
```
tomas-pluskal: It seems a bit strange to use System.load() and then catch the UnsatisfiedLinkException. Perhaps this whole logic should be moved to the NativeLibraryLoader class.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,349 @@
...
+      System.load("glpk_4_60");
+      System.load("glpk_4_60_java");
+    } catch (UnsatisfiedLinkError e) {
+      try {
+        // GLPK requires two libraries
+        String[] libs = {"glpk_4_60", "glpk_4_60_java"};
+        NativeLibraryLoader.loadLibraryFromJar("glpk-4.60", libs);
+      } catch (Exception e1) {
+        throw new RuntimeException(e1);
```
tomas-pluskal: Use MSDKRuntimeException instead

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,349 @@
...
+    this.numberOfCandidates = numberOfCandidates;
+    this.constraints = constraints;
+    this.deviation = new Deviation(deviation);
+  }
+
+  /**
+   * <p> Class ConstraintsGenerator. </p>
+   * This class allows to construct a Sirius object FormulaConstraints using MolecularFormulaRange object
+    */
+  public static class ConstraintsGenerator {
```
tomas-pluskal: Is there any advantage in having this class as a static class inside SiriusIdentificationMethod? I would change it to a stand-alone class, it will make SiriusIdentificationMethod shorter and more clear.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,349 @@
...
+  public int getNumberOfCandidates() {
+    return numberOfCandidates;
+  }
+
+  /**
+   *
+   * <p> Method for reading custom MS files (just 2 columns with mz & intentsity values </p>
+   * Does similar to mgf parser functionality
+   */
+  public static MsSpectrum readCustomMsFile(File file, String delimeter)
```
tomas-pluskal: This does not belong here. If you want to have an import function for such files, you should create a corresponding io module. But it seems to me this does the same as the MSP import module that is already there: https://github.com/msdk/msdk/blob/master/msdk-io-msp/src/main/java/io/github/msdk/io/msp/MspImportAlgorithm.java

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,349 @@
...
+  }
+
+  /**
+   * <p> Method for processing spectra by Sirius module </p>
+   * Transformation of MSDK data structures into Sirius structures and processing by Sirius
+   * Method is left to be protected for test coverage
+   */
+  protected List<IdentificationResult> siriusProcessSpectra() throws MSDKException {
+    Spectrum<Peak> siriusMs2;
+    MsSpectrum msdkSpectrumMs2 = ms2.remove(0);
```
tomas-pluskal: This looks strange - why are you modifying the contents of ms2?

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,349 @@
...
+        MutableMs2Spectrum siriusMs2 = new MutableMs2Spectrum(peaks);
+        ms2spectra.add(siriusMs2);
+      }
+    }
+  }
+
+  @Nullable
+  @Override
+  public Float getFinishedPercentage() {
+    return null;
```
tomas-pluskal: Is there any way to estimate the progress of the whole operation?

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,349 @@
...
+
+    return new SimpleMsSpectrum(mz, intensity, size, type);
+  }
+
+  /**
+   * <p> Method for processing spectra by Sirius module </p>
+   * Transformation of MSDK data structures into Sirius structures and processing by Sirius
+   * Method is left to be protected for test coverage
+   */
+  protected List<IdentificationResult> siriusProcessSpectra() throws MSDKException {
```
tomas-pluskal: This method never checks if the operation was cancelled

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
evgerher: First I thought about making it as one of the methods of SiriusIdentificationMethod class, but decided that it will be too overloaded.
As it is strongly correlated with Sirius module, I left it here.

But it is not a problem, I can move it.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
tomas-pluskal: I think you should not modify the original ms2 list here. Because the user who is using the module might not expect that the ms2 list (which the user gave to you as a parameter), will be modified during the process. It is better if you create a copy of the list for your own use, then you can do anything you want with it (the copy of the list will contain the references to the same spectra as the original list, there is no need to create a copy of each spectrum).

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
tomas-pluskal: Shorter classes are always better.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
evgerher: Oh, really, thank you for advice

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
evgerher: I decided to remove the old code and it is now fully controlled by native library loader.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
evgerher: Sirius module does not have this opportunity.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
evgerher: I can't cancel the sirius processing job, the only possibility is to cancel before it is started.
Job takes a big amount of time, the preparation stuff does not, I am not sure that cancellation will have sense.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
```diff
@@ -0,0 +1,349 @@
...
+
+    double mz[] = msdkSpectrumMs2.getMzValues();
+    double intensity[] = LocalArrayUtil.convertToDoubles(msdkSpectrumMs2.getIntensityValues());
+
+    siriusMs2 = sirius.wrapSpectrum(mz, intensity);
+    String ionization = ion.getName();
+    PrecursorIonType precursor = sirius.getPrecursorIonType(ionization);
+
+    /* MutableMs2Experiment allows to specify additional fields and it is exactly what comes from .getMs2Experiment */
+    MutableMs2Experiment experiment = (MutableMs2Experiment) sirius.getMs2Experiment(parentMass, precursor, null, siriusMs2);
```
tomas-pluskal: Here you should not assume that sirius.getMs2Experiment() always returns MutableMs2Experiment. Maybe future Sirius versions will return a different object. What is the actual return type of the function?
You should first check using the instanceof operator, before doing the cast to MutableMs2Experiment.

msdk-id-sirius/src/main/java/SiriusIdentificationMethod.java
evgerher: Well, it does quite similar stuff as the msp import module, but the files' structure differs.
This method does not have any specific structure.

This method allows to test on not-structured files.

tomas-pluskal:
I just noticed that none of your classes have a package declaration (i.e., they belong to default package). It is common practice to put classes into packages organized by domain name. For this module, you can make a package io.github.msdk.id.sirius and move all the classes there (both main classes and test classes). Check other modules if anything is unclear:
https://github.com/msdk/msdk/blob/master/msdk-io-txt/src/main/java/io/github/msdk/io/txt/TxtImportAlgorithm.java#L14

msdk-id-sirius/src/main/java/io/github/msdk/id/sirius/ConstraintsGenerator.java
```diff
@@ -79,7 +79,7 @@ public FormulaConstraints generateConstraint(MolecularFormulaRange range) {
       if (max!= maxNumberOfOneElements) constraints.setUpperbound(element, max);
     }
 
-    logger.debug("ConstraintsGenerator finished");
+    logger.debug("io.github.msdk.id.sirius.ConstraintsGenerator finished");
```
tomas-pluskal: It seems like the package name was added to more places than it was intended?

msdk-id-sirius/src/test/resources/simplelogger.properties
```diff
@@ -4,8 +4,8 @@ org.slf4j.simpleLogger.defaultLogLevel=info
 # Must be one of ("trace", "debug", "info", "warn", or "error").
 # If not specified, the default logging detail level is used.
 #org.slf4j.simpleLogger.log.xxxxx=
-org.slf4j.simpleLogger.logger.SiriusMs2Test=trace
-org.slf4j.simpleLogger.logger.NativeLibraryLoader=trace
+org.slf4j.simpleLogger.logger.io.github.msdk.id.sirius.SiriusMs2Test=trace
```
tomas-pluskal: This also looks wrong..

msdk-id-sirius/src/main/java/io/github/msdk/id/sirius/ConstraintsGenerator.java
evgerher: Oh sorry, autorefactor by IDEA and I forgot to check this

chhh: When you do auto-refactoring IDEA splits the refactorings into categories, like "actual code changes", "changes in comments", "uncategorized text changes" etc. Always take a look at everything that's not code, or just straight reject those (right click, exclude)

msdk-id-sirius/pom.xml
```diff
@@ -124,17 +190,67 @@
...
-    <dependency>
-      <groupId>org.openscience.cdk</groupId>
-      <artifactId>cdk-silent</artifactId>
-      <version>${cdk.version}</version>
-    </dependency>
+
     <dependency>
       <groupId>io.github.msdk</groupId>
       <artifactId>msdk-io-txt</artifactId>
       <version>0.0.12-SNAPSHOT</version>
```
tomas-pluskal: Use ${project.version} instead of specifying the version number here

msdk-id-sirius/pom.xml
```diff
@@ -10,26 +10,26 @@
...
+      <name>Test repository</name>
+      <url>https://bio.informatik.uni-jena.de/repository/libs-release-oss/</url>
     </repository>
+    <!--<repository>-->
+      <!--<id>maven-main-repo</id>-->
+      <!--<name>Maven main repository</name>-->
+      <!--<url>http://repo.maven.apache.org/maven2/</url>-->
+    <!--</repository>-->
     <repository>
       <id>ebi</id>
```
tomas-pluskal: What is the EBI repo used for?

evgerher: Removed it

msdk-id-sirius/src/main/java/io/github/msdk/id/sirius/FingerIdWebMethod.java
```diff
@@ -0,0 +1,646 @@
...
+
+
+  /**
+   * <p>Constructor for a FingerIdWebMethod</p>
+   * @param experiment - Ms2Experiment that was used to get ionAnnotation
+   * @param ionAnnotation - SiriusIonAnnotation returned from SiriusIdentificationMethod with FTree field specified
+   * @param candidatesAmount - amount of candidates to be returned from this method
+   * @throws MSDKException if any
+   */
+  public FingerIdWebMethod(Ms2Experiment experiment, SiriusIonAnnotation ionAnnotation, int candidatesAmount) throws MSDKException {
```
tomas-pluskal: Please add @Nullable or @Nonnull annotations to all parameters of public methods. Also, it is better to use Integer instead of int for the parameters.

msdk-id-sirius/src/main/java/io/github/msdk/id/sirius/FingerIdWebMethod.java
```diff
@@ -0,0 +1,646 @@
...
+   * <p>Method returns URIBuilder for the Web server</p>
+   * @param path API functionality you are interested in
+   * @return builder
+   * @throws URISyntaxException if any
+   */
+  private URIBuilder getFingerIdURI(String path) throws URISyntaxException {
+    if (path == null)
+      path = "";
+    URIBuilder builder = new URIBuilder(FINGERID_SOURCE);
+    builder.setPath("/csi-fingerid-" + FingerIdWebMethod.fingerIdVersion() + path);
```
tomas-pluskal: Why not use the constant FINGERID_VERSION instead of calling a private method fingerIdVersion()?

evgerher: Old idea, I used to think that it will be nice to add in future a capability of autochecking the version of FingerId
Removed.

PR #265, opened 2018-08-13
Hot fix the size of spectrum read by MGF import module

evgerher:
There was a bug that MgfSpectrum.
It used to count less m/z-intensity pairs than it was in reality. 
Because of mistake tests also did not notice it.


tomas-pluskal:
It seems the travis build fails due to some nullpointerexceptions. Can you take a look at it?

PR #268, opened 2018-09-14
Fingerid fix

evgerher:
Reduce the size of big requests.
Get only top N MsSpectrum for experiment. Order is done by largest intensity value from mz/intensity pairs.
Get only top M pairs of each MsSpectrum. Order is done by Intensity.


msdk-id-sirius/pom.xml
```diff
@@ -244,6 +244,11 @@
       <artifactId>cdk-silent</artifactId>
       <version>${cdk.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.github.msdk</groupId>
+      <artifactId>msdk-io-mgf</artifactId>
+      <version>0.0.15-SNAPSHOT</version>
```
tomas-pluskal: should use ${project.version} here

PR #269, opened 2018-09-25
Code refactor

evgerher:
Refactored code, moved some methods to MsSpectrumUtil, deleted class SpectrumPreprocessor.


evgerher:
I suppose let it exist until I will create more fixes.
I wish to remove JUL for sirius dependencies and use slf4j-jul-bridge.
https://www.slf4j.org/legacy.html#jul-to-slf4j

tomas-pluskal:
I think you don't need to care too much about the logging dependency in SIRIUS. In my view, this is a poor design issue on their side and hopefully they will address it in some future release.

PR #271, opened 2019-01-15
New isotope incorporation module

susannefuerst:
Hi all,

this is my first contribution to the MSDK. I implemented some methods to simulate not only natural 
isotope patterns but also patterns that will be expected/produced from tracing experiments. I also implemented a method to correct intensities for natural abundancies [1].

I am quite new to Maven, so please have a look if I have done it right. There have been some SIRIUS issues, when I run the install without the skipTests option. So I used the skipTest option but I'm not sure if this was a good idea.

For more detailed information on the code see below.

Looking forward to your comments.

Best,
Susanne

###################   HOW TO USE THE CODE   ########################

The most important classes for this are:

**IsotopePatternSimulaterRequest.class
IsotopePatternSimulator.class
IsotopePatternSimulatorResponse.class**

Parameters have to be set in the **IsotopePatternSimulatorRequest.class**. Assume you labeled with 
(15N-amino)(13C)5-glutamine and you want to simulate the isotope pattern of the glutamate pool 
where 60% of the pool consists of (15N)(13C)5-glutamate. Then your Fragment.class is given by the 
formula C5H9NO4 and the capacity C5N (for capacity definition see the **Fragment.class** documentation). Set the IncorporationRate to 0.6 and use the **IsotopePatternSimulator#simulate** method for simulation (this means you don't expect an independent tracer incorporation).

If you want to simulate a glutamate pool where 10% consist of 15N-glutamate, 20% consist of
(13C)5-glutamate and 50% consist of (15N)(13C)5-glutamate (so you expect an independent tracer incorporation) use the **IsotopePatternSimulator#simulateIndependentTracerIncorporation** method. You need to set tracer1 (C), tracer2 (N) and the related incorporations on the IsotopePatternSimulaterRequest.class. Samples can be found  in the **IncorporationMapCorrectionTest**.

When you set analyseMassShifts = true in the IsotopePatternSimulatorRequest.class you will get
the information which isotopes induced certain peaks, but this could cost a lot.

If you want to determine the isotope incorporation in your experiment you have to correct for the natural abundancies. Therefore I implemented the ideas from [1].

The most important class for this is the **IncorporationMap.class** and it's **#correctForNaturalAbundance** method. See the documentation of this class for more information. Samples can be found in the I**ncorporationMapCorrectionTest.class**.

For validation of the algorithms I compared my simulation of(natural) patterns with the simulation method of natural patterns published by [2] (see **Element.class#multiElementSpectrum**). I also first simulated certain pattern and subsequently used the natural abundance correction to check if I obtain the simulated incorporation (see **IncorporationMapCorrectionTest.class**).

#################################   REFS   ######################################

[1]: Carreer, W. J.; Flight, R. M. & Moseley, H. N. B. A Computational Framework for High-Throughput Isotopic Natural Abundance Correction of Omics-Level Ultra-High Resolution FT-MS Datasets. Metabolites, 2013 , 3 

[2]: Yergey, James A.: A General Approach to Calculating Isotopic Distributions For Mass Spectrometry. In: International Journal of Mass Spectrometry 52(1983), 2/3, S. 337-349. - ISSN 1387-3806


tomas-pluskal:
Hi Susanne, 

Thanks for the submission! I looked through the code briefly. 

It seems your code is now designed pretty much as an independent library on its own, with its own data model etc. If you want to keep the code as it is, I honestly don't see too much benefit of integrating it into MSDK (neither for you nor for us). If, however, you are willing to adapt the code to MSDK "style", then it could be useful, because it would allow people to use it together with other methods in MSDK. 

The main things that need changing are:

- We have a general interface for MSDK methods (https://github.com/msdk/msdk/blob/master/msdk-datamodel/src/main/java/io/github/msdk/MSDKMethod.java) that your method(s) should implement

- Use MSDK data model (https://github.com/msdk/msdk/tree/master/msdk-datamodel/src/main/java/io/github/msdk/datamodel). For example, your module uses its own MassSpectrum class, while MSDK uses the general MsSpectrum interface, which creates confusion.

- MSDK generally uses CDK modules for working with isotopes, see for example https://github.com/msdk/msdk/blob/master/msdk-isotopes-isotopepattern/src/main/java/io/github/msdk/isotopes/isotopepattern/IsotopePatternGeneratorAlgorithm, which is using the IsotopePatternGenerator from CDK ( https://github.com/cdk/cdk/blob/master/tool/formula/src/main/java/org/openscience/cdk/formula/IsotopePatternGenerator.java)

I don't want to have duplicated functionality in the library, so we should consider whether some parts of your library might better fit into CDK. Isotope work is of course somewhat borderline - some of it is pure chemistry (i.e., fits better into CDK), while things like isotope tracing may fit better into MSDK, because they are really only used for mass spectrometry. Looking at your module, though, I have the impression that it basically extends the current functionality of CDK's IsotopePatternGenerator, doesn't it?

susannefuerst:
Hi Tomas,

thanks for your feedback. Yes if you think it could be useful I would like to adapt the code to MSDK style. I mean, I have to check the effort. Maybe it makes more sense to simply rewrite the code using MSDK style. I could create a new branch and do frequent pull request so that you can check the process., 

Maybe I can create an MSDK algorithm that uses the CDK IsotopePatternGenerator to simulate the tracing. What do you think? If you feel I should better commit the tracing simulation to CDK do you have any advice for me how to proceed? should I first contact someone responsible or just create a first pull request?

What about the natural abundance correction? I think this would need a new datatype (something like the IncorporationMap in my code). Should I create a new datatype in the datamodel module? What would the be the right place for the actual correction algorithm?

Regards,
Susanne

tomas-pluskal:
Hi Susanne,

I think your work fits better into MSDK than into CDK, because it is really mass spectrometry oriented (for isotope tracing applications, specifically). 

If I understand correctly, the input of your module is a chemical formula plus additional parameters (labeling definition), and the output is a predicted isotope distribution (= predicted mass spectrum). In that case, I think the best strategy might be to expand/rewrite the existing MSDK IsotopePatternGenerator module to add the functions that you implemented (https://github.com/msdk/msdk/blob/master/msdk-isotopes-isotopepattern/src/main/java/io/github/msdk/isotopes/isotopepattern/IsotopePatternGeneratorAlgorithm). What do you think about this approach?

susannefuerst:
Hi Tomas,

Yes you are right about the input and the output. You need to know the formula of the molecule/fragment , the used tracer, the amount of tracer that can be incorporated into the molecule/fragment and the incorporation rate(s) you want to simulate. If you would like to have a more precise explanation on how this should work please let me know.

I think what I want to do fits to the IsotopePatternGeneratorAlgorithm. I would like to implement a method that returns an IsotopePattern, that includes the compositions of the peak inducing components (similar to the MZmine datatype https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/datamodel/impl/ExtendedIsotopePattern.java).  Should I implement such a MsSpectrum within the datamodel module or within the isotopepattern module?

It seems to make sense to extend the #generateIsotopes() method within the IsotopePatternGeneratorAlgorithm to already return such an IsotopePattern (maybe using a flag). Then I can reuse this method to simulate the traced spectra. What do you think?

tomas-pluskal:
Hi Susanne,

My recommended strategy would be to 

1) Create your own class (e.g. TracingIsotopePattern) that implements IsotopePattern interface but also has additional methods as needed. You can place this class in the msdk/isotopes/isotopepattern package.

2) Create a new TracingIsotopePatternGeneratorAlgorithm in the same package, that will be analogous to the IsotopePatternGeneratorAlgorithm but will contain all the logic related to tracing experiments. It might be easier for users to have these functions separated from the existing IsotopePatternGeneratorAlgorithm (I generally prefer to have shorter classes with well-defined scope of functionality). I am not sure if TracingIsotopePatternGeneratorAlgorithm should extend IsotopePatternGeneratorAlgorithm or it should simply call its generateIsotopes() function or if these classes should have their own, completely independent implementations. You can decide as you develop the code.

On a side note, the CDK's IsotopePatternGenerator can already provide the composition of each isotope (if you set storeFormula==true), so you can take advantage of that.

susannefuerst:
Hi Tomas,

Okay I will do so.

susannefuerst:
Hi Tomas,

yesterday I had a conversation with my PI about my intention to contribute my code to the msdk. He adviced me to first proceed with my research before I make any contributions. Therefore I wanted to let you know that I will postpone my final commit to a later point in time. 

Maybe I will find some interesting MZmine TODOs in the meantime :-)

robinschmid:
You are always welcome to contribute and collaborate :)

Cheers
Robin

tomas-pluskal:
Hi Susanne, 
Ok no problem. Please let me know if you decide to contribute something in the future, and we can coordinate our efforts.
Cheers,
Tomas

